PR #46, opened 2013-06-12
Check also if obj is not None

volpino:
We encountered this issue when restoring a database backup. At that moment the database had different data than the harvest object queue and the whole fetch_consumer failed because obj.source raised an exception (as obj was None).

This little fix solves the problem on 1.8, I don't know if also the 2.0 is affected.

### DIFF
--- a/ckanext/harvest/queue.py
+++ b/ckanext/harvest/queue.py
@@ -117,6 +117,8 @@ def fetch_callback(message_data,message):
 
         try:
             obj = HarvestObject.get(id)
+            if obj is None:
+                raise Exception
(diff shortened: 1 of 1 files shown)


metaodi:
@amercader I think this PR is no longer relevant, but probably it wouldn't hurt to merge it either :wink:

davidread:
On master it already does the test for obj.

But this PR is for the 1.8 release branch, which is a bit curious. Might as well merge it I guess.

davidread:
We don't merge onto release branches, so closing.

PR #67, opened 2013-09-16
add deletion to ckanharvester for harvesting < 2.0

joetsoi:
maybe this should be a config option as well

Prior to 2.0, https://github.com/okfn/ckan/pull/545 the api returns a 403 error. Deletions will appear in the list of revisions in the gather stage but will error during the import stage.


davidread:
No work on this for 12 months, to address the issues, so closing.

We're about to do merge large changes to the ckan harvester too. In addition I'm not sure there are any CKAN 1.x instances out there anyway. Feel free to ressurrect if still needed.

PR #107, opened 2014-09-10
Job reporting fixes

Zharktas:
Job reporting has some inconsistencies which causes confusing behavior. 

If harvest job is executed and there are zero new revision, the harvest job never receives job finished timestamp.

If harvest job is executed again, it refresher all harvested datasets, but since it creates new HarvestObjects, they don't have report_status saved and the job report shows that all datasets are deleted although nothing actually changed.

### DIFF
--- a/ckanext/harvest/harvesters/base.py
+++ b/ckanext/harvest/harvesters/base.py
@@ -178,8 +178,6 @@ class HarvesterBase(SingletonPlugin):
 
                 else:
                     log.info('Package with GUID %s not updated, skipping...' % harvest_object.guid)
-                    harvest_object.report_status = 'not modified'
-                    harvest_object.save()
                     return
 
                 # Flag the other objects linking to this package as not current anymore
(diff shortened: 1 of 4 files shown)


ckanext/harvest/harvesters/base.py
```diff
@@ -178,6 +178,8 @@ def _create_or_update_package(self, package_dict, harvest_object):
 
                 else:
                     log.info('Package with GUID %s not updated, skipping...' % harvest_object.guid)
+                    harvest_object.report_status = 'not modified'
```
davidread: I'd much rather report_status was only updated in queue.py, rather than introduce a new place where it potentially gets updated. Also, this is a deprecated method to be updating.

My preference would be for fetch_stage() to return True, False or "unchanged" to queue.py and set report_status there.

ckanext/harvest/logic/action/update.py
```diff
@@ -317,6 +317,8 @@ def harvest_jobs_run(context,data_dict):
                           .first()
                     if last_object:
                         job_obj.finished = last_object.import_finished
+                    else:
+                        job_obj.finished = job['gather_finished']
```
davidread: Whatever we do about the rest, this 2 lines are good.

Zharktas:
I modified it based on your comments, now import stage returns "unchanged" if it the dataset is unchanged and queue.py set report_status based on this.

davidread:
I've taken these changes, added some tests and related fixes and created PR #177 as a replacement for this one.

Issue #114, opened 2014-12-08
remove references to obsolete ResourceGroup


florianm:
Using ckanext-harvest master on CKAN master (2.3a), clearing imported datasets from a harvest job will fall over now obsolete references to `ResourceGroup`.

`logic/action/update.py`l116ff contains [references](https://github.com/ckan/ckanext-harvest/blob/623daa72c4a5b0f0f3f4796f1dbed44049b34043/ckanext/harvest/logic/action/update.py#L116) to the now obsolete ckan model ResourceGroup.

Could be implemented along [these lines](https://github.com/ckan/ckanext-qa/pull/8/files) by toggling the lines referencing db table `resource_group` depending on whether ResourceGroup can be imported.

davidread:
This was merged, closing.

PR #118, opened 2014-12-12
Removed ResourceGroup from query when using CKAN 2.3 or above

clementmouchet:
PR to fix issue https://github.com/ckan/ckanext-harvest/issues/114


clementmouchet:
Finally took time to fix this one. It will delete resource views, resource revisions & resources in CKAN 2.3 and above

davidread:
Reads fine to me. @amercader do you want to double check?

davidread:
Reading it again and it still makes sense. Merging.

PR #119, opened 2014-12-16
Make sure new packages get a unique 'name'

### LINKED ISSUE #117: That URL is already in use (reopen)
Same issue, not solved yet:
https://github.com/ckan/ckanext-harvest/issues/116

The code is from this repo.

metaodi:
This fixes #117.

The problem is that a "new" package could have a 'name' attributes that's already taken. By using `gen_new_name()` we can be sure the name is valid for this instance. The conditional is only there to ensure that the existing 'name' is used if possible, but maybe that's not even necessary.

### DIFF
--- a/ckanext/harvest/harvesters/base.py
+++ b/ckanext/harvest/harvesters/base.py
@@ -202,7 +202,7 @@ class HarvesterBase(SingletonPlugin):
                 context.pop('__auth_audit', None)
 
                 # Set name for new package to prevent name conflict, see issue #117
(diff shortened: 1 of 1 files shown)


metaodi:
@amercader I guess you have lot to do with CKAN 2.3, but when can we expect this PR to be merged?

amercader:
@metaodi I'll look at this tomorrow

amercader:
@metaodi thanks for the patch

PR #121, opened 2015-01-13
Fetch remote organization via action api

### LINKED ISSUE #120: Creating remote orgs does not work
From the ckan-dev mailing list:

> Hello all,
>
> I'm trying to import datasets from a remote CKAN 2.2 instance, and I would like to import owner organizations as well. According to the harvester docs, it should be possible (and fairly easy) to do so using the remote_orgs:"create" configuration. However, after doing some tests, no organization is being imported.
>
> Looking at the code, the remote organizations are being fetched using the /group route of the v2 of the API. But testing the api/2/rest/group/ with Postman, I can see it only returns a list of ids from groups, not organizations.
>
> I believe that's the reason why the harvester can't find the owner_org. 
> Is this a known bug? Should api/2/rest/group/ return organization ids too or is what I'm getting expected behavior?
>
> In case the latter is true, should we change CKANHarvester to use v3 of the api and /organization_show ? Or is there a better workaround?
>
> Thanks for your help in advance.
> Cheers,
>
> Rodrigo Parra

I could verify this behaviour. A call to /api/2/rest/group used to return organization as well, but now this is not the case anymore so the creating of remote organizations is broken.

metaodi:
Organizations used to be returned by `/api/2/rest/group`, this is what the
old implementation used to fetch the information to create the remote
organization on the local instance of CKAN.

With this commit the Action API is used to fetch the same information.

Fixes #120

### DIFF
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -5,6 +5,7 @@ from ckan import model
 from ckan.model import Session, Package
 from ckan.logic import ValidationError, NotFound, get_action
 from ckan.lib.helpers import json
+from ckan.lib.munge import munge_name
 
 from ckanext.harvest.model import HarvestJob, HarvestObject, HarvestGatherError, \
                                     HarvestObjectError
@@ -51,7 +52,7 @@ class CKANHarvester(HarvesterBase):
         return http_response.read()
 
     def _get_group(self, base_url, group_name):
-        url = base_url + self._get_rest_api_offset() + '/group/' + group_name
+        url = base_url + self._get_rest_api_offset() + '/group/' + munge_name(group_name)
         try:
             content = self._get_content(url)
             return json.loads(content)
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -47,7 +47,7 @@ class CKANHarvester(HarvesterBase):
         except urllib2.URLError, e:
             raise ContentFetchError(
                 'Could not fetch url: %s, error: %s' % 
-                (url, e.reason)
+                (url, str(e))
             )
         return http_response.read()
 
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -44,10 +44,10 @@ class CKANHarvester(HarvesterBase):
 
         try:
             http_response = urllib2.urlopen(http_request)
-        except urllib2.HTTPError, e:
+        except urllib2.URLError, e:
             raise ContentFetchError(
-                'Could not fetch url: %s, HTTP error code: %s' % 
-                (url, e.code)
+                'Could not fetch url: %s, error: %s' % 
+                (url, e.reason)
             )
         return http_response.read()
 
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -337,7 +337,14 @@ class CKANHarvester(HarvesterBase):
                         log.info('Organization %s is not available' % remote_org)
                         if remote_orgs == 'create':
                             try:
-                                org = self._get_organization(harvest_object.source.url, remote_org)
+                                try:
+                                    org = self._get_organization(harvest_object.source.url, remote_org)
+                                except:
+                                    # fallback if remote CKAN exposes organizations as groups
+                                    # this especially targets older versions of CKAN
+                                    log.debug('_get_organization() failed, now trying _get_group()')
+                                    org = self._get_group(harvest_object.source.url, remote_org)
+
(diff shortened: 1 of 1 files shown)


ckanext/harvest/harvesters/ckanharvester.py
```diff
@@ -324,7 +337,14 @@ def import_stage(self,harvest_object):
                         log.info('Organization %s is not available' % remote_org)
                         if remote_orgs == 'create':
                             try:
-                                org = self._get_group(harvest_object.source.url, remote_org)
+                                try:
+                                    org = self._get_organization(harvest_object.source.url, remote_org)
+                                except:
```
amercader: @metaodi this is bad, see comment above

ckanext/harvest/harvesters/ckanharvester.py
```diff
@@ -48,6 +52,15 @@ def _get_group(self, base_url, group_name):
         except Exception, e:
             raise e
 
+    def _get_organization(self, base_url, org_name):
+        url = base_url + self._get_action_api_offset() + '/organization_show?id=' + org_name
+        try:
+            content = self._get_content(url)
+            content_dict = json.loads(content)
+            return content_dict['result']
+        except Exception, e:
```
amercader: @metaodi this should be avoided, as it will catch Exceptions that are not related (eg if the json parsing failed). I guess that if we are targeting an old CKAN instance we will get a 404 if the action is not available, so we should only catch that (probably a urllib2 execption)

metaodi: I understand the reasoning, but here it really doesn't matter why it fails. The implementation is analog to the implementation of `_get_group`, but of course it could be changed in both places. If - as you wrote - the parsing of the JSON failed, it should fail. Maybe it's better to remove the try/catch entirely, and delegate the exception handling to the caller.

In this particular case the [caller catches all exception in an `except` block](https://github.com/metaodi/ckanext-harvest/blob/120-create-remote-orgs/ckanext/harvest/harvesters/ckanharvester.py#L353), as any exception simply means, that the remote organization could not be fetched.

What do you think?

davidread: There are also lots of programming errors that you can make, that cause different exceptions, which are nothing to do with the server's response. Not swallowing those exceptions helps fix the code. Hence why we have a CKAN policy of making the effort to identify the correct exception in except clauses. It's also mentioned in PEP8:

> When catching exceptions, mention specific exceptions whenever possible instead of using a bare except: clause.

davidread:
PS Cheers for submitting this - most helpful

ckanext/harvest/harvesters/ckanharvester.py
amercader: I hadn't noticed the line on 353. I still think that catch all exceptions are bad (even if wrote the `_get_group` one a long time ago :speak_no_evil: ) but I agree that we can let this one go in.

amercader: But then again @davidread is right and we shouldn't do it, so let's fix `_get_group` and `_get_organization` to be more explicit :)

metaodi:
@davidread @amercader okay, I tried to clean this all up. I introduced two now *Error classes, one to wrap the urllib2 exception and one to wrap either the HTTP or the json parse error. I replaced all generic `except:` blocks and almost all generic `Exception` usages in the CKANHarvester.

During my tests I also found an [issue with the remote_groups](https://github.com/metaodi/ckanext-harvest/commit/935b9dda0100325028b01eb45a363b5730670c16), if you want me to create a separate PR for that, just let me know.

ckanext/harvest/harvesters/ckanharvester.py
```diff
@@ -36,17 +41,34 @@ def _get_content(self, url):
...
             http_request.add_header('Authorization',api_key)
-        http_response = urllib2.urlopen(http_request)
 
+        try:
+            http_response = urllib2.urlopen(http_request)
+        except urllib2.HTTPError, e:
+            raise ContentFetchError(
+                'Could not fetch url: %s, HTTP error code: %s' % 
+                (url, e.code)
+            )
```
davidread: I believe this also needs to catch urllib2.URLError, in cases such as where the url has bad syntax, or the socket connection times out.

davidread:
Thanks @metaodi - I think that's another really valuable improvement

ckanext/harvest/harvesters/ckanharvester.py
metaodi: Good catch, I'll replace it with urllib2.URLError (which is the super class of urllib2.HTTPError), this should cover the exceptions at this point.

davidread:
I didn't know about HTTPError inheriting URLError - cheers!

metaodi:
@davidread a thanks for this hint, replaced it with str() as you suggested.

davidread:
Thanks @metaodi

PR #123, opened 2015-01-22
Use the organization slug to get remote org

### LINKED ISSUE #122: Use organization slug to get remote org
Currently the harvester uses the `owner_org` value to get a organization via the API of a remote CKAN instance. This value is sometimes the slug of an organization, but could also be its guid. This guid cannot be used to get an organization via the API.

By always using the slug, this problem can be avoided.

metaodi:
The API action organization_show does currently not respond to calls
using the GUID of an organization, therefore the slug name has to be
used. It is available in the package_dict and can be used for the call
to the remote API.

Fixes #122

### DIFF
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -330,12 +330,13 @@ class CKANHarvester(HarvesterBase):
                 # Assign dataset to the source organization
                 package_dict['owner_org'] = local_org
             else:
-                if not 'owner_org' in package_dict:
-                    package_dict['owner_org'] = None
-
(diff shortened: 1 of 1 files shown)


rossjones:
Which version of CKAN is this for?

For me http://master.ckan.org/api/3/action/organization_show?id=ef3f1cf3-d144-4a57-9bca-9f437ac34e0b and http://master.ckan.org/api/3/action/organization_show?id=123-org returns the same thing.

metaodi:
@rossjones you're right, sorry I think I was too fast here. Maybe the problem is somewhere else. I'll check back with the [original reporter](https://lists.okfn.org/pipermail/ckan-dev/2015-January/008565.html).

PR #134, opened 2015-06-10
Clear command

metaodi:
Clear is an often used command, that is currently only available in the web UI. I think it would be very useful to have this possibility on the command line as well.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -115,6 +115,9 @@ The following operations can be run from the command line using the
       harvester rmsource {id}
         - remove (inactivate) a harvester source
 
+      harvester clearsource {id}
+        - clear a harvester source (delete all associtated datasets)
+
       harvester sources [all]
         - lists harvest sources
           If 'all' is defined, it also shows the Inactive sources
--- a/README.rst
+++ b/README.rst
@@ -113,10 +113,10 @@ The following operations can be run from the command line using the
         - create new harvest source
 
       harvester rmsource {id}
-        - remove (inactivate) a harvester source
+        - remove (deactivate) a harvester source, whilst leaving any related datasets, jobs and objects
 
       harvester clearsource {id}
-        - clear a harvester source (delete all associtated datasets)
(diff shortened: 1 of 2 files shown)


davidread:
Looks good! Should be really useful to have this plumbed in.

README.rst
```diff
@@ -115,6 +115,9 @@ The following operations can be run from the command line using the
       harvester rmsource {id}
         - remove (inactivate) a harvester source
 
+      harvester clearsource {id}
+        - clear a harvester source (delete all associtated datasets)
```
davidread: Typo "associtated"

ckanext/harvest/commands/harvester.py
```diff
@@ -20,6 +20,9 @@ class Harvester(CkanCommand):
       harvester rmsource {id}
         - remove (inactivate) a harvester source
 
+      harvester clearsource {id}
+        - clear a harvester source (delete all associtated datasets)
```
davidread: Typo again.

davidread: I think it would be good to be clearer and use the language in the logic function here (and in the README): "Clears all datasets, jobs and objects related to a harvest source, but keeps the source itself."

ckanext/harvest/commands/harvester.py
```diff
@@ -20,6 +20,9 @@ class Harvester(CkanCommand):
       harvester rmsource {id}
         - remove (inactivate) a harvester source
```
davidread: This could be improved to be: "remove (deactivate) a harvester source, whilst leaving any related datasets, jobs and objects."

davidread:
Do consider these wording changes. I'll leave Adria to merge since he's responsible for harvest.

metaodi:
@davidread thanks for the review, I updated the wording accordingly.

amercader:
This is great, thanks @metaodi for the patch and @davidread for reviewing

Issue #138, opened 2015-06-24
Change timezone of displayed dates


metaodi:
Is there a way to easily adjust the offset in hours for all the datetimes that the harvester show in the frontend?
As far as I know, all dates are saved with UTC timezone and also displayed that way. This regularly leads to confusing for admins that are not aware of this (we're currently at UTC+2).

Is there an easy fix for that?

davidread:
I don't know of a timezone setting. It'd definitely be useful, ckan-wide, if you'd like to add it. I'm pretty sure that's why dates are all saved as UTC.

Harvesting is one key place, you're right. Also displaying Last Updated / Created fields and in Activity Stream.

metaodi:
It would be nice to have this setting ckan-wide, I just added this issue here, as this is currently the one place where it bothers our users the most. Harvesters are normally time-controlled and hence they'd like to know when happend what, also to cross-reference with web server logfiles that might be in the local time.

davidread:
I've added it in core ckan too: https://github.com/ckan/ckan/issues/2494. Please do contribute it if you have a need - shouldn't be hard to add it to the central helper function that displays dates.

davidread:
This was merged, so closing the issue.

PR #142, opened 2015-07-09
 [WIP] [#138] Add data attributes to support timezone conversion

metaodi:
This PR is based on PR https://github.com/ckan/ckan/pull/2505 and fixes issue #138.

It uses data attributes to embed ISO-formatted dates, which then in turn can be converted to the correct timezone of the user visiting the page.


metaodi:
@davidread this was a WIP PR :wink:  but yeah it doesn't really hurt that it is merged already, but it will only work in combincation with https://github.com/ckan/ckan/pull/2505

PR #144, opened 2015-07-14
An extension point to allow filtering packages during harvesting

filipefigcorreia:
This allows others to create customized CKAN harvester to exclude packages using any metadata element as criteria. To use this extension point, one just needs to create a new plugin class that inherits from `CKANHarvester` and overrides the `info()` and `_should_import()` methods.

### DIFF
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -273,6 +273,9 @@ class CKANHarvester(HarvesterBase):
                 log.warn('Remote dataset is a harvest source, ignoring...')
                 return True
 
+            if not self._should_import(package_dict):
+                package_dict['state'] = 'deleted'
+
             # Set default tags if needed
             default_tags = self.config.get('default_tags',[])
             if default_tags:
@@ -433,6 +436,11 @@ class CKANHarvester(HarvesterBase):
                     harvest_object, 'Import')
         except Exception, e:
             self._save_object_error('%r'%e,harvest_object,'Import')
+
+    def _should_import(self,package_dict):
+        # Override in harvesters that inherit from CKANHarvester to avoid importing unwanted packages.
+        # Overriding classes should do their own logging of the reason the package is not imported.
+        return True
 
 class ContentFetchError(Exception):
     pass


metaodi:
Maybe it would make sense to generalize this further by adding such an extension point to the `BaseHarvester`. It doesn't seem to be CKAN specific. What do you think?

filipefigcorreia:
@metaodi Sounds good, I can give it a go, it will probably involve adding a new method to [`IHarvester`](https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/interfaces.py#L3) and changing `fetch_and_import_stages()` to invoke the filter method. Do you agree?

metaodi:
Hm, I wouldn't change `IHarvester`, this change seems too small for this. I think you should move your code the HarvesterBase class, and call your method in [`_create_or_update_package`](https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/harvesters/base.py#L149)

filipefigcorreia:
That could work, although there's quite a bit of logic in [`import_stage()`](https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/harvesters/ckanharvester.py#L254) before [`_create_or_update_package()`](https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/harvesters/base.py#L149) gets called, which we could just bypass if filtering is done before the [`import_stage()`](https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/harvesters/ckanharvester.py#L254) gets called, in [`fetch_and_import_stages()`](https://github.com/ckan/ckanext-harvest/blob/fd93a69da9cad71896e21f82cbb21024f7629c34/ckanext/harvest/queue.py#L310). So, changing `IHarvester` can turn out to be more efficient. 

An additional thought is that the kind of filtering that I mean to do is somewhat similar to [this validation](https://github.com/ParadigmaXis/ckanext-harvest/blob/77e33bde719cde630cdc26da1824c60b2f407ef5/ckanext/harvest/harvesters/ckanharvester.py#L272) made by `import_stage()`, so it may make sense to use the same mechanism for both filterings. But I haven't given it enough thought yet.

amercader:
Now that #158 is merged I'd prefer that this was implemented as a configuration option that allowed to pass a search filter to be applied on the search to get the remote datasets, so you don't get them at all in the first place. Happy to help with a spec if someone wants to take it on.

Issue #145, opened 2015-07-20
Catch "There are no new harvesting jobs" exception


metaodi:
Based on [an email to the ckan-dev mailinglist](https://lists.okfn.org/pipermail/ckan-dev/2015-July/009128.html), I'm adding this issue to catch the ugly "There are no new harvesting jobs" exception when running the paster command. I think a simple error message will do the trick.

PR follows.

metaodi:
Here is the PR: #146

metaodi:
Fixed with #146

PR #146, opened 2015-07-20
[#145] Throw + catch a custom exception if there are no jobs to run

metaodi:
If there are no harvesting jobs to run, there was always an ugly
exception message when using the paster command. This replaces the ugly
output with a proper message and uses a custom exception to allow others
to deal with this error differently.


amercader:
Thanks for the patch @metaodi

PR #148, opened 2015-07-22
Catch, record and recover from temporary db problems

davidread:
Not quite sure why we get these db exceptions - maybe due to upgrading and forgetting to restart fetch & gather queues. 

Anyhow, according to our logs we saw this happen on 8 days in the past 6 months, which is useful to track. And I can't remember, but think this will not recover from itself without this remove() (or manually restarting the fetch & gather processes).

### DIFF
--- a/ckanext/harvest/queue.py
+++ b/ckanext/harvest/queue.py
@@ -209,8 +209,17 @@ def gather_callback(channel, method, header, body):
     # Get a publisher for the fetch queue
     publisher = get_fetch_publisher()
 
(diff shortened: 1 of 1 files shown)


davidread:
I'll make it more specific catch

davidread:
@amercader I've changed the exception catch as suggested. Would you mind reviewing?

davidread:
That'll merge now @amercader

davidread:
Conflicts fixed again... @amercader

PR #152, opened 2015-08-19
Fix #105 by adapting harvest_source_clear to removal of authz models

### LINKED ISSUE #105: SQL query to clean harvester
Hello,

I have 2 questions about SQL queries used to clean (I see those queries in the file  ckanext/harvest/logic/action/update.py).

Why do you delete in this order  "package_role" and then "user_object_role" ?
In my case, the sub-query 

select user_object_role_id from package_role 

returns more than 5 000 000 entities and I want to clean only 1700 datasets.
If we just invert the deletion, like this

delete from user_object_role where id in (select user_object_role_id from package_role where package_id in (select id from package where state = 'to_delete' )) and context = 'Package';
delete from package_role where package_id in (select id from package where state = 'to_delete' );

the result will be faster?!?!

My other question is about the last query :

delete from package where id in (select id from package where state = 'to_delete');

why don't you simply do like this :

delete from package where state = 'to_delete';

best regards
Antoine

florianm:
## TL;DR
Fix #105 by upgrading harvest_source_clear to not delete from authz models removed in migration 078 for CKAN versions > 2.4.

## Details
Migration [078](https://github.com/ckan/ckan/blob/63e3bc9186f02a4bf1cce24b4c5d937e696c04c3/ckan/migration/versions/078_remove_old_authz_model.py) removes the legacy authz models, including `package_role` and `user_object_role`.
These two models are deleted from in `harvest_source_clear`. This PR moves the SQL queries to delete from the two corresponding tables to a conditional block which only runs for CKAN pre-2.5.

## Assumptions
* Migration 078 applies to versions >= 2.5. Maintainers: is that correct?
* The models `package_role` and `user_obkect_role` exist in CKAN up to version 2.4.

## Effect
* Harvest > View a harvesting source > Admin > Clear will work again.
* Works in CKAN 2.5, but untested in CKAN < 2.5

### DIFF
--- a/ckanext/harvest/logic/action/update.py
+++ b/ckanext/harvest/logic/action/update.py
@@ -141,7 +141,7 @@ def harvest_source_clear(context,data_dict):
         (select id from package where state = 'to_delete');
         '''
     # CKAN pre-2.5: authz models were removed in migration 078
-    if toolkit.check_ckan_version(max_version='2.4'):
+    if toolkit.check_ckan_version(max_version='2.4.99'):
         sql += '''
         delete from user_object_role where id not in 
         (select user_object_role_id from package_role) and context = 'Package';
--- a/ckanext/harvest/logic/action/update.py
+++ b/ckanext/harvest/logic/action/update.py
@@ -140,14 +140,22 @@ def harvest_source_clear(context,data_dict):
         delete from resource_group where package_id  in 
         (select id from package where state = 'to_delete');
         '''
+    # CKAN pre-2.5: authz models were removed in migration 078
+    if toolkit.check_ckan_version(max_version='2.4'):
+        sql += '''
+        delete from user_object_role where id not in 
+        (select user_object_role_id from package_role) and context = 'Package';
+        delete from package_role where package_id in 
+        (select id from package where state = 'to_delete');
+        '''
+
+
     sql += '''
     delete from harvest_object_error where harvest_object_id in (select id from harvest_object where harvest_source_id = '{harvest_source_id}');
     delete from harvest_object_extra where harvest_object_id in (select id from harvest_object where harvest_source_id = '{harvest_source_id}');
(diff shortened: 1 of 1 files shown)


ckanext/harvest/logic/action/update.py
```diff
@@ -140,14 +140,22 @@ def harvest_source_clear(context,data_dict):
         delete from resource_group where package_id  in 
         (select id from package where state = 'to_delete');
         '''
+    # CKAN pre-2.5: authz models were removed in migration 078
+    if toolkit.check_ckan_version(max_version='2.4'):
```
davidread: This should be '2.4.99' to include patch releases too

davidread:
Looks good! Just that one improvement requested before we merge.

davidread:
> *  Migration 078 applies to versions >= 2.5. Maintainers: is that correct?
> *  The models package_role and user_obkect_role exist in CKAN up to version 2.4.
Yes those assumptions are correct.

florianm:
@davidread thanks for the feedback! Let's see what Travis thinks of my patch.

davidread:
I'm keen to merge this, but one more thing: why did you swap over the order of deleting user_object_role and package_role? There is a foreign key relation between them, so in the absence of testing it on an old ckan, it's probably safest to keep it in the order it was, don't you think?

Here's the constraint and the way I see it you need to delete the package_role before any user_object_role that it references:

    package_role_table = Table('package_role', meta.metadata,
           Column('user_object_role_id', types.UnicodeText, ForeignKey('user_object_role.id'), primary_key=True),

florianm:
That makes sense. I probably wasn't careful enough - I'm spoiled by django
orm and south migrations - will swap them as you suggested.

davidread:
No problems, I swapped em and merged on a derived branch datawagovau-105-clean-the-cleaner

florianm:
Thanks Dave! Sorry, I got sidetracked and didn't get the  change in in time.

PR #158, opened 2015-09-10
CKAN harvester to use v3 package_search API instead of v2 package_list/revisions

rossjones:
Rather than using the revisions in v2 API this now uses the package_search API so that we can extend it with proper filters in future.

Currently fails on package_update with 

```
2015-09-10 17:56:32,723 ERROR [ckanext.harvest.harvesters.base] (IntegrityError) insert or update on table "resource" violates foreign key constraint "resource_revision_id_fkey"
DETAIL:  Key (revision_id)=(6c00105f-c4f0-457e-be6f-ac2c4d550edd) is not present in table "revision".
```


ckanext/harvest/harvesters/ckanharvester.py
```diff
@@ -374,7 +366,7 @@ def import_stage(self,harvest_object):
             # Find any extras whose values are not strings and try to convert
             # them to strings, as non-string extras are not allowed anymore in
             # CKAN 2.0.
-            for key in package_dict['extras'].keys():
+            for key in package_dict.get('extras', {}).keys():
```
amercader: `extras` is a list of dicts, so calling `.keys()` won't work. We need to iterate of them and check the `key` value.
The logic on the lines below is also broken for v3 dicts

davidread: Am working on this.

davidread:
Could really do with this harvesting from older CKANs too, so doing the RESTful interface as a fall-back.

ckanext/harvest/harvesters/ckanharvester.py
rossjones: You mean check the value value ;) 

Is the intention that default_extras are always added, but only override existing values if override_extras=True?

rossjones:
By older CKAN's you mean pre 2.0?

ckanext/harvest/harvesters/ckanharvester.py
```diff
@@ -246,8 +236,10 @@ def fetch_stage(self,harvest_object):
                                         (url, e),harvest_object)
             return None
 
+        content = json.loads(content)['result']
```
amercader: This looks fragile, we shouldn't assume it will be a CKAN API response (could be a 404, etc)

rossjones: Should probably handle non 200 responses in _get_content above.

davidread:
Ok it basically works now in the tests, although I've not tried it against a real ckan to check, especially a 2nd harvest, which was broken before.

Would be good to add in the support for still harvesting using the restful calls as a fallback, @rossjones if you have a chance?

amercader:
Looking really good!

One thing that could massively boost the performance is storing the data_dict on the gather stage. As we are now using `package_search` which will return it for us we might as well store it as the HarvestObject content and skip completely the fetch stage (saving N requests to the remote CKAN, where N is the number of datasets).

This can be done on a separate PR, but we should probably paginate more cleverly, in batches of 100 and allowing more than 1000 remote datasets.

amercader:
BTW the RDF harvester on ckanext-dcat does the same thing of storing the contents on gather and skip fetch:

https://github.com/ckan/ckanext-dcat/blob/master/ckanext/dcat/harvesters/rdf.py#L206:L221

davidread:
Now I've done:
* use package_search for the first harvest (in addition to subsequent ones, which uses the modification time)
* package_search paging, 100 items at a time (takes 3 seconds per page for DGU so ok I think)
* merged organization filter stuff but using package_search instead

Could do with some real-world testing as I've been relying on my tests with mock CKAN.

davidread:
BTW I looked for CKANs pre v2.0 and couldn't find any, so I don't think there is need to retain the old API calls as a fallback after all.

davidread:
Ross's `IntegrityError` mentioned in his first comment is to do with revision_id in a resource. This is now tested and fixed.

davidread:
I harvested 25000 datasets of data.gov.uk and it took:
* 8 minutes to get them all from data.gov.uk
* 2 minutes to create the harvest objects
* 5 hours to create the datasets (including adding them to SOLR)

davidread:
When paginating, could do it by date, or do a bit of overlap of the offsets to ensure none are missed.

wardi:
@davidread or use a > last_seen_modified_date in the query instead of offset

amercader:
@davidread is this WIP or can it be reviewed?

davidread:
This is the status:

* [x] convert to v3 API
* ~~option to harvest from CKANs via old RESTful API~~ - decided this is not necessary
* [ ] ensure when paging through that no records are missed (e.g. using some overlap)
* [ ] more real-world testing

davidread:
I'm just doing some more testing now harvesting a selection of ckans.

davidread:
Hooray! Ready for review now @amercader

amercader:
This is fanstastic @davidread, thanks a lot for your work on this. It makes harvests much faster and efficient and allows for implementing further filtering on the results from the remote CKAN. I just added a minor picky change (9dfeb154ebb5d360a120fe7abaaa77213da4771c).

:clap: :space_invader:

davidread:
Cool, thanks @amercader

PR #159, opened 2015-09-30
Added the posibility of harvesting only datasets belonging to concrete organizations

memaldi:
Added the posibility of harvesting only datasets belonging to concrete organizations through the config key:

```
{"limit_org": ["organization1-id", "organization2-id", "organizationN-id"]}
```

### DIFF
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -46,7 +46,7 @@ class CKANHarvester(HarvesterBase):
             http_response = urllib2.urlopen(http_request)
         except urllib2.URLError, e:
             raise ContentFetchError(
-                'Could not fetch url: %s, error: %s' % 
+                'Could not fetch url: %s, error: %s' %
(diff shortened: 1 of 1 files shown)


davidread:
I think this is really useful - thanks for submitting this @memaldi . Actually my colleague @rossjones was also going to submit a PR for this feature, based on this commit:

https://github.com/datagovuk/ckanext-harvest/commit/01fdbbf682c007a38e98c06065a48b5b8addbe65

TBH I prefer the name 'organizations_filter_include' / 'organizations_filter_exclude'. Your solution by filtering in the gather_stage seems more efficient, as you avoid doing package_show on the excluded datasets. However I think it could be slightly neater if organization_pkg_ids was a set, rather than list.

What do you think?

memaldi:
I agree, we could move @rossjones's code to the gather_stage, it's ok for me.

davidread:
Closing this in favour of https://github.com/ckan/ckanext-harvest/pull/168 which takes this and adds tests and documentation.

PR #163, opened 2015-10-20
CKAN Harvester option to include/exclude organizations [WIP]

davidread:
Also modernized tests and factories.


davidread:
I've broken this up into 4 PRs

PR #164, opened 2015-10-21
purge_queues improvements

davidread:
* purge_queues command now has warning about impact of Redis flushall
* add some (log) output when you run a purge.


amercader:
Wow, I didn't realise we were wiping all Redis keys! This is not great. AFAICT is not trivial to just delete the `harvest*` keys so at least we should use `flushdb` to just delete the keys on the current database. If different CKAN installs share the same Redis queue you must use different databases with `ckan.harvest.mq.redis_db` to avoid clashes. I added a commit to address this

PR #167, opened 2015-10-21
Add framework for testing harvesters

davidread:
No-one has written tests for any harvesters AFAICT. This PR makes it easy, and avoids messing with queues! It provides a run_harvest(url, harvester) method which tests can call. That runs a harvest with the same job states, error handling and report_status as if you ran a harvest normally using the queues. It returns lots of info about the harvest that it's just done.

Some ckanharvester tests are also provided, which tests it against a mock CKAN.

The factories are much improved and existing tests modernized a bit.

There are a couple of benign changes to harvest_source_show, model setup and queue gather_callback.

### DIFF
--- a/ckanext/harvest/logic/action/get.py
+++ b/ckanext/harvest/logic/action/get.py
@@ -30,10 +30,22 @@ def harvest_source_show(context,data_dict):
     :param id: the id or name of the harvest source
     :type id: string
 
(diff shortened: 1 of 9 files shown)


amercader:
Great stuff @davidread, thanks.

davidread:
Cheers @amercader

PR #168, opened 2015-10-21
CKAN Harvester option to include/exclude organizations

davidread:
NB this is based on testing framework #167 which needs merging first


davidread:
Closed in favour of #169

Issue #106, opened 2014-09-03
Delete faulty HarvestJobs to avoid stalling fetch and gather queues


rodowi:
The fetch queue fails on start when a buggy harvest job is pending, preventing other jobs to be completed.

```
$ sudo service supervisord restart
Deteniendo supervisord:                                    [  OK  ]
Iniciando supervisord:                                     [  OK  ]
$ sudo supervisorctl
ckan_fetch_consumer BACKOFF    Exited too quickly (process log may have details)
ckan_gather_consumer RUNNING    pid 15856, uptime 0:00:10
supervisor> status
```

```
sqlalchemy.exc.InvalidRequestError: Can't attach instance <Package at 0x3d8a550>; another instance with key (<class 'ckan.model.package.Package'>, (u'ac046ba9-3aaf-4da5-b8a7-998a860719a1',)) is already present in this session.
```
_one of many errors that might happen_

It seems that the only exit might be to remove the harvest source and reharvest, but this has undesired consequences, such as removing all the previous datasets from the source, which is unacceptable for production sites.

So I'm trying to explore the best way to remove faulty harvest jobs. This is what I got so far:

1. [I could delete rows holding that job_id in the Database](https://github.com/ckan/ckanext-harvest/blob/33000dbfbc12236ad43aa36eb2ad0298df3aeaf3/ckanext/harvest/model/__init__.py#L168)
2. I could remove the job through ckanext.harvest.model.HarvestJob but I haven't been able to fetch a Job. I'm getting `sqlalchemy.exc.InvalidRequestError: SQL expression, column, or mapped entity expected - got '<class 'ckanext.harvest.model.HarvestJob'>'`
3. We could add a "rmjob" option to the cli.

Is any one trying to fix this?

davidread:
#170 provides job_abort

PR #170, opened 2015-10-28
Add run_test, job_abort, source commands

davidread:
* run_test - for running a whole harvest on the command-line
* job_abort - for aborting a limbo job
* source - for showing a single harvest source
* allowing a source to be specified by name in several commands


davidread:
Fixes #106

PR #171, opened 2015-10-28
Add extension to CheeseShop / PyPI

### LINKED ISSUE #149: Add extension to CheeseShop / PyPI
Wondering why this extension isn't available via the `CheeseShop`. Would _massively_ ease the usage in project requirements.

davidread:
This PR are the docs and version increment. If @amercader is happy then I'll actually put it on PyPI and add amercader and iward as owners.

Fixes #149

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -606,6 +606,60 @@ guidelines that apply to CKAN core, described in
 `CONTRIBUTING <https://github.com/ckan/ckan/blob/master/CONTRIBUTING.rst>`_.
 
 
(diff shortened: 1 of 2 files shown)


amercader:
Happy with this, but the extension can't be pip installed right? Because the requirements are on `requirements.txt`

davidread:
Ok, not I have the will to go to the effort of doing regular releases and writing changelogs, and the original instigator admits it's not a big deal after all.

PR #174, opened 2015-10-29
Changed harvest source URL validator to allow duplicate URLs with unique configs and modified to make PEP-8 compliant.

LondonAppDev:


davidread:
Perhaps you could change this PR title/description to reflect the change in
functionality in the other commit

LondonAppDev:
Hey David,

Sorry, the change in the second commit is not finished yet. I was hoping to
submit the PEP-8 changes and then make a separate request for the unique
URL change. In hindsight I should have used a different branch for my
second change.

Cheers,
Mark

On Thu, Oct 29, 2015 at 7:34 PM David Read <notifications@github.com> wrote:

> Perhaps you could change this PR title/description to reflect the change in
> functionality in the other commit
>
> On Thursday, 29 October 2015, LondonAppDev <notifications@github.com>
> wrote:
>
> > ------------------------------
> > You can view, comment on, or merge this pull request online at:
> >
> > https://github.com/ckan/ckanext-harvest/pull/174
> > Commit Summary
> >
> > - Modified to make PEP-8 compliant.
> >
> > File Changes
> >
> > - *M* ckanext/harvest/logic/validators.py
> > <https://github.com/ckan/ckanext-harvest/pull/174/files#diff-0> (73)
> >
> > Patch Links:
> >
> > - https://github.com/ckan/ckanext-harvest/pull/174.patch
> > - https://github.com/ckan/ckanext-harvest/pull/174.diff
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/ckan/ckanext-harvest/pull/174>.
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ckan/ckanext-harvest/pull/174#issuecomment-152296422>.
>
-- 
*Mark Winterbottom | *Full-Stack Software Engineer
Mobile: +44 (0) 7918 713 807
Website: http://mark.winterbottom.me
LinkedIn: https://www.linkedin.com/in/markwinterbottom
Blog: http://londonappdeveloper.com

LondonAppDev:
Hi @davidread,

I have modified the title to reflect the newest changes I've made. Previously the URL for each harvest source would need to be unique.

Using an OAI-PMH harvester I ran into the challenge where I had multiple organisations that used the same OAI-PMH endpoint but with a unique configuration (pointing to a specific 'set' per org).

To me it makes sense to allow the same URL to be re-used as long as the configuration is unique. 

Kind regards,
Mark

ckanext/harvest/logic/action/update.py
```diff
@@ -428,18 +448,23 @@ def harvest_source_reindex(context, data_dict):
...
-    log.debug('Updating search index for harvest source {0}'.format(harvest_source_id))
+    package_dict = logic.get_action('harvest_source_show')(
+        context, {'id': harvest_source_id})
+    log.debug('Updating search index for harvest source {0}'.format(
+        harvest_source_id))
 
     # Remove configuration values
     new_dict = {}
-    if package_dict.get('config'):
+    if package_dict.get('config', None):
```
davidread: I'm not sure adding `, None` adds anything here since that is the default - perhaps you might remove it? There are some others too that would be usefully removed.

ckanext/harvest/logic/action/delete.py
```diff
@@ -5,30 +5,23 @@
 
 log = logging.getLogger(__name__)
 
+
 def harvest_source_delete(context, data_dict):
-    '''
+    """
```
davidread: Actually the CKAN convention is for single quotes i.e. `'''` if you don't mind

davidread:
I'm basically in favour of the premise of this PR - allowing a new harvest source of the same URL as along as they have a different config.

It seems a little bit of a kludge in `harvest_source_url_validator` to look at the harvest source to see what the config is, when we've already got the package object. @amercader what do you think?

LondonAppDev:
Apologies, didn't know convention was single quotes for docstrings. I've also updated the ', None' calls in my code and some other calls that were like that.

davidread:
Nice one! Would you mind also merging from master? There are some conflicts.

ckanext/harvest/logic/action/delete.py
```diff
@@ -5,30 +5,21 @@
...
 def harvest_source_delete(context, data_dict):
-    '''
-    Deletes an existing harvest source
+    '''Deletes an existing harvest source.
 
     This method just proxies the request to package_delete,
     which will delete the actual harvest type dataset and the
     HarvestSource object (via the after_delete extension point).
-
-    :param id: the name or id of the harvest source to delete
```
davidread: This is correct isn't it? Not sure why you're deleting it. You're right to delete the `:returns:` though.

ckanext/harvest/logic/action/update.py
```diff
@@ -239,18 +273,19 @@ def harvest_objects_import(context, data_dict):
...
-    package_id_or_name = data_dict.get('package_id',None)
+    source_id = data_dict.get('source_id',)
+    harvest_object_id = data_dict.get('harvest_object_id')
+    package_id_or_name = data_dict.get('package_id')
 
-    segments = context.get('segments',None)
+    segments = context.get('segments')
 
-    join_datasets = context.get('join_datasets',True)
+    join_datasets = context.get('join_datasets', rue)
```
davidread: oops!

ckanext/harvest/logic/action/delete.py
LondonAppDev: I deleted the params because it doesn't accept an ID. I should have added `context` and `data_dict` as params.

ckanext/harvest/logic/action/update.py
```diff
@@ -518,9 +560,13 @@ def harvest_source_reindex(context, data_dict):
     if package_dict.get('config'):
         config = json.loads(package_dict['config'])
         for key, value in package_dict.iteritems():
-            if key not in config:
-                new_dict[key] = value
+            if value:
```
davidread: this seems like a duplicate test from the next line. remove this one?

ckanext/harvest/logic/action/update.py
```diff
@@ -518,9 +560,13 @@ def harvest_source_reindex(context, data_dict):
...
+                if value and key not in config:
```
davidread: why did you the `if value`? Why shouldn't you delete the key for a blank value?

ckanext/harvest/logic/action/delete.py
davidread: Ah, got you. You would be completely right for normal python, but this is a CKAN convention for logic functions. The params are essentially the `data_dict`, and we use the docstring to document them, instead of context & data_dict themselves. More examples here: http://docs.ckan.org/en/latest/api/#ckan.logic.action.get.member_list but I completely agree that this is a non-obvious departure from python norms.

ckanext/harvest/logic/action/update.py
```diff
@@ -239,18 +273,19 @@ def harvest_objects_import(context, data_dict):
...
-    package_id_or_name = data_dict.get('package_id',None)
+    source_id = data_dict.get('source_id',)
+    harvest_object_id = data_dict.get('harvest_object_id')
+    package_id_or_name = data_dict.get('package_id')
 
-    segments = context.get('segments',None)
+    segments = context.get('segments')
 
-    join_datasets = context.get('join_datasets',True)
+    join_datasets = context.get('join_datasets')
```
davidread: The default value still needs to be True though

ckanext/harvest/logic/action/update.py
LondonAppDev: the duplicated `if value` is my mistake (shouldn't rush changes, lesson learnt). The `if value` fixed the issue of deleting harvest endpoints which had a config (my comment mentioned here: https://github.com/ckan/ckanext-harvest/commit/58a873ac7a31e7bfc956967d4e6f49368e3799e7#commitcomment-14129955). However I see now that, what I thought was a good fix, could break other things. I need to look for a solution to the `revision_timestamp` and `tracking_summary` values having missing objects.

LondonAppDev:
@davidread I have undone my fix for the bug in ckan/ckanext-harvest which causes 500 error when deleting harvest sources, as my fix breaks other things. I've updated the title to reflect this. 

I need more time to fix the delete sources bug so I'll do this in a separate commit. In the mean time this pull request will allow adding the same URL multiple times with different configs.

davidread:
I've processed the PEP8 and other style changes now and created a new PR with just the validator changes: https://github.com/ckan/ckanext-harvest/pull/184

Issue #180, opened 2015-11-03
modified date was not checked properly when re-harvesting CKAN datasets


davidread:
When updating a dataset it checks that the metadata_modified value of the remote dataset is newer than when it was last harvested. However a bug caused the check to always pass, meaning the opportunity to skip the dataset was never taken.

The problem was the package_show was called with a bad context - it specified `schema=default_create_package_schema()` (needed for when calling package_create/update later) but it caused the package_show to not include the metadata_modified field.

PR #181, opened 2015-11-03
Fix modified date check

davidread:
Issue: #180


davidread:
The problem only affects CKAN 2.3 onwards.

PR #184, opened 2015-11-03
Allow Harvest Sources with duplicate URLs if they have unique configs

davidread:
This is #174 from @LondonAppDev only without the PEP8 changes, for clarity.

@LondonAppDev said
> Previously the URL for each harvest source would need to be unique.
> 
> Using an OAI-PMH harvester I ran into the challenge where I had multiple organisations that used the same OAI-PMH endpoint but with a unique configuration (pointing to a specific 'set' per org).

I said:
> I'm basically in favour of the premise of this PR

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -31,24 +31,27 @@ Installation
 
       ckan.harvest.mq.type = amqp
 
+2. Activate your CKAN virtual environment, for example::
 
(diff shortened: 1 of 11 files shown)


davidread:
I think this is fine, merging

PR #186, opened 2015-11-04
Dataset count mismatch bug fix WIP

davidread:
Fix from https://github.com/GSA/ckanext-harvest/commit/1cb44485922aa7

I'm not clear that this is a bug, but the fix looks fine! We should add a test anyway.


metaodi:
@davidread is this still valid and/or WIP?

davidread:
Sounds like this was to cope with a gremlin in the data that GSA suffered. No evidence others are having the same problem, so I'll close it.

PR #187, opened 2015-11-04
Show changes for every dataset when viewing a job WIP

davidread:
Adapted from https://github.com/ckan/ckanext-harvest/compare/stable...GSA:2.3-upgrade#diff-225a47b85c61888dadfbf974f109bb56R44

I've not tried this out yet. Concerned about speed, so defaults to off for now.


metaodi:
@davidread is this still valid and/or WIP?

davidread:
I thought it might be useful to cherry-pick this from the US's work, but I've not got round to work on the issues I was concerned about. Happy to close this.

PR #188, opened 2015-11-11
New API function harvest_source_patch

metaodi:
Add a `harvest_source_patch` API function to be able to update a harvest source with only the changes you want to make. Internally this is a proxy to the already existing `package_patch`. This means it is only available in CKAN >= 2.3, in older versions `harvest_source_update` is used instead.

### DIFF
--- a/ckanext/harvest/logic/action/patch.py
+++ b/ckanext/harvest/logic/action/patch.py
@@ -1,0 +1,58 @@
+'''API functions for partial updates of existing data in CKAN'''
+
+import logging
+from ckan.logic import get_action
+from ckanext.harvest.plugin import DATASET_TYPE_NAME
+
+log = logging.getLogger(__name__)
+
+
+def harvest_source_patch(context, data_dict):
+    '''
+    Patch an existing harvest source
+
+    This method just proxies the request to package_patch,
+    which will update a harvest_source dataset type and the
+    HarvestSource object. All auth checks and validation will
+    be done there. We only make sure to set the dataset type.
+
+    Note that the harvest source type (ckan, waf, csw, etc)
+    is now set via the source_type field.
+
+    All fields that are not provided, will be stay as they were before.
+
+    :param id: the name or id of the harvest source to update
+    :type id: string
+    :param url: the URL for the harvest source
+    :type url: string
+    :param name: the name of the new harvest source, must be between 2 and 100
+        characters long and contain only lowercase alphanumeric characters
+    :type name: string
+    :param title: the title of the dataset (optional, default: same as
+        ``name``)
+    :type title: string
+    :param notes: a description of the harvest source (optional)
+    :type notes: string
+    :param source_type: the harvester type for this source. This must be one
+        of the registerd harvesters, eg 'ckan', 'csw', etc.
+    :type source_type: string
+    :param frequency: the frequency in wich this harvester should run. See
+        ``ckanext.harvest.model`` source for possible values. Default is
+        'MANUAL'
+    :type frequency: string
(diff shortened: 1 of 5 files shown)


ckanext/harvest/tests/test_action.py
```diff
@@ -307,6 +307,58 @@ def test_update(self):
...
+
+    @classmethod
+    def setup_class(cls):
+
+        cls.action = 'harvest_source_patch'
+
+        super(TestHarvestSourceActionPatch, cls).setup_class()
+
+        # Create a source to udpate
+        source_dict = cls.default_source_dict
```
davidread: It's really great to have this improvement. On CKAN tests, we've changed our style a bit over the past couple of years - see http://docs.ckan.org/en/latest/contributing/testing.html#guidelines-for-writing-new-style-tests and are not keen on creating fixtures that are shared between tests like this.

To help with this, you can quickly setup fixtures like a HarvestSource in each test using the new factories. See example here:
https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/tests/test_action.py#L311-L316

ckanext/harvest/logic/action/patch.py
```diff
@@ -0,0 +1,58 @@
...
+
+    :returns: the updated harvest source
+    :rtype: dictionary
+    '''
+    log.info('Patch harvest source: %r', data_dict)
+
+    data_dict['type'] = DATASET_TYPE_NAME
+
+    context['extras_as_string'] = True
+    source = get_action('package_patch')(context, data_dict)
```
davidread: Ah, the tests are spotting that package_patch was introduced in CKAN 2.3, so I guess you'll need to check the CKAN version, I guess here, but particularly in the tests (call SkipTest).

metaodi:
@davidread thanks for your quick review, I refactored the tests to match the new guidlines, especially regarding the fixtures. I'd appreciate it, if you could look over my changes, now that I'm finished and all the tests pass again.

davidread:
Cheers for moving the common fixtures out of those tests :+1: 

I'm new to `nost.tools.ok_`. Seems to make the line 2 chars shorter but exception more long winded. Are there any other advantages?
```
>>> ok_(4 == 5)
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/vagrant/ckan/local/lib/python2.7/site-packages/nose/tools/trivial.py", line 22, in ok_
    raise AssertionError(msg)
AssertionError: None
>>> assert 4 == 5
Traceback (most recent call last):
  File "<console>", line 1, in <module>
AssertionError
```

metaodi:
@davidread I'd say the main advantage is that you have the option to add a custom message just like with all the other assertion method. With `assert` all you get is an `AssertionError`, that's it. I think in the end I didn't use it somewhere, but it might be a help to debug.

And to quote the [nose documentation](http://nose.readthedocs.org/en/latest/testing_tools.html) :wink: :
> ***nose.tools.ok_(expr, msg=None)***
Shorthand for assert. *Saves 3 whole characters!*

davidread:
Custom message like these? I can't see much difference. Not very important though...!
```
>>> ok_(1 == 2, 'not the same')
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/vagrant/ckan/local/lib/python2.7/site-packages/nose/tools/trivial.py", line 22, in ok_
    raise AssertionError(msg)
AssertionError: not the same

>>> assert 1 == 2, 'not the same'
Traceback (most recent call last):
  File "<console>", line 1, in <module>
AssertionError: not the same
```

davidread:
Cheers @metaodi

PR #190, opened 2015-11-13
Metaodi harvest source patch

davidread:
This is based on https://github.com/ckan/ckanext-harvest/pull/188 but test changes from @davidread


metaodi:
@davidread thanks for picking this up, I had quite some struggles with the test code.

PR #192, opened 2015-11-17
Make sure all status keys are always returned

metaodi:
It makes it harded to parse the API response if you always have to check if 'deleted' is set and if what value it has. I think to simply return 0 for all status values is good practice.


davidread:
Agreed :+1:

PR #193, opened 2015-11-19
Catch the correct exception to not miss gather errors

metaodi:
The current code expects a HTTPError, but `_get_content` swallows this
low-level exception and instead raises a ContentFetchException.

By adding another custom exception it's possible to distinigish between
a general error and a "not found" error (HTTP code 400).

### DIFF
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -45,6 +45,17 @@ class CKANHarvester(HarvesterBase):
 
         try:
             http_response = urllib2.urlopen(http_request)
+        except urllib2.HTTPError, e:
+            if e.getcode() == 400:
+                raise ContentNotFoundError(
+                    'Content not found at url: %s, error %s' %
+                    (url, str(e))
+                )
+            else:
+                raise ContentFetchError(
+                    'Could not fetch url: %s, error: %s' %
+                    (url, str(e))
+                )
         except urllib2.URLError, e:
             raise ContentFetchError(
                 'Could not fetch url: %s, error: %s' %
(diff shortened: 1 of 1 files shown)


davidread:
So, in this try block, CKAN asks for the revisions since the last harvest and then asks for the details of each revision. It's making a requests like these:

http://beta.ckan.org/api/search/revision?since_time=2016-10-05T19:42:45.854533
http://beta.ckan.org/api/rest/revision/f74eb786-763c-4ead-898c-89023fbdb422

Agreed that the existing catch for 400 error is doing nothing now.

The current logic is that if there is any error we should try the other method of harvesting instead. You're saying that it should only try the other way of harvesting if there is a 400 error. However I'm not clear that 400 does mean not found, or that all other errors should avoid the second harvesting method.

I would have thought that the only time you were sure there was no point trying another way of harvesting is if there had been no revisions, which is a 404 error.

e.g. http://beta.ckan.org/api/search/revision?since_time=2017-10-05

So I think you need to change it from 400 to 404 and reverse your logic at the bit line 222.

Or have I missed something?

metaodi:
I actually didn't check the error code, I just used the same as it was before (and I think always was), but I can certainly change that.

I don't see that I need to reverse my logic on line 222. If I get the 400/404 error, I raise a `ContentNotFoundError` exception, which is then catched on line 222, `get_all_packages` is set to `True` and the program continues.

davidread:
Following your logic, 400/404 means there were no revisions, why then should it try and harvest all datasets?

I think you're trying to resurrect long-since broken code...

metaodi:
I have to check this a little further. Because what I actually want is to receive a gather error if there was an HTTP error. Maybe we can get rid of this whole revision-checking-thingy.

davidread:
I think this code harks back to a time when some versions of ckan didn't have a revision API. However there are some other products which also don't have a revision API, such as DKAN. So the idea is that in this case we don't have a gather error, it just continues to the fallback method, which is to get the full list of packages. If we were to save a gather error then I think people would be confused why it still managed to harvest ok. We really need a save_gather_warning here...

I've put my suggested change here: https://github.com/ckan/ckanext-harvest/pull/198/files just to be clear.

metaodi:
I close this PR in favor of @davidread PR #198.

PR #197, opened 2015-11-23
Use the ckan version of munge_tag if available, but provide a fallback

davidread:
(fallback is for older ckans)

### DIFF
--- a/ckanext/harvest/harvesters/base.py
+++ b/ckanext/harvest/harvesters/base.py
@@ -21,14 +21,14 @@ from ckanext.harvest.model import HarvestJob, HarvestObject, HarvestGatherError,
 from ckan.plugins.core import SingletonPlugin, implements
 from ckanext.harvest.interfaces import IHarvester
 
-try:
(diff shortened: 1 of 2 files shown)


ckanext/harvest/harvesters/base.py
```diff
@@ -21,16 +21,29 @@
 from ckan.plugins.core import SingletonPlugin, implements
 from ckanext.harvest.interfaces import IHarvester
 
+if p.toolkit.check_ckan_version(min_version='2.3'):
+    from ckan.lib.munge import munge_tag
```
metaodi: Is there a reason not to use a `try` ... `except ImportError` block for this? This would work, even if a future version of CKAN does no longer include `munge_tag`.

davidread: I did do that initially, but the problem turned out to be that although ckan 2.2 does have this function, it is a rubbish version and fails a couple of the tests e.g. turning 'river quality' into 'riverquality'.

If we ever were to take the function out of ckan then we can add a max_version here without too much trouble.

PR #194, opened 2015-11-20
Raise general exceptions

raphaelstolt:
Also raise ContentFetchError for general exceptions or sub its exceptions (i.e.`httplib.BadStatusLine`).

PR #198, opened 2015-11-23
Catch exceptions from urllib2.urlopen more comprehensively

davidread:
I think '400' errors were from something as old as CKAN 0.6 - ignore now these aren't in service any more.

### DIFF
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -45,7 +45,10 @@ class CKANHarvester(HarvesterBase):
         try:
             http_response = urllib2.urlopen(http_request)
         except urllib2.HTTPError, e:
-            raise ContentFetchError('HTTP error: %s' % e.code)
+            if e.getcode() == 404:
+                raise ContentNotFoundError('HTTP error: %s' % e.code)
+            else:
+                raise ContentFetchError('HTTP error: %s' % e.code)
         except urllib2.URLError, e:
             raise ContentFetchError('URL error: %s' % e.reason)
         except httplib.HTTPException, e:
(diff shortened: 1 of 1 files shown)


raphaelstolt:
Will this PR also be merged into branch `release-v2.0`?

davidread:
@raphaelstolt Yes that would make sense, as it is a bit of a bug not handling well invalid harvest source URLs.

What do you think of the PR?

raphaelstolt:
@davidread Looks :+1: to me.

ckanext/harvest/harvesters/ckanharvester.py
```diff
@@ -35,21 +36,23 @@ def _get_search_api_offset(self):
...
+                raise ContentNotFoundError('HTTP error: %s' % e.code)
+            else:
+                raise ContentFetchError('HTTP error: %s' % e.code)
         except urllib2.URLError, e:
-            raise ContentFetchError(
-                'Could not fetch url: %s, error: %s' %
-                (url, str(e))
-            )
+            raise ContentFetchError('URL error: %s' % e.reason)
+        except httplib.HTTPException, e:
```
metaodi: Where is `httplib` used? If I [read the documentation](https://docs.python.org/2/library/httplib.html#httplib.HTTPException) correctly, this exception is only raised from their module. Or is `urllib2` based on `httplib` somehow?

metaodi:
@davidread I think this PR goes in the right direction, much better then my #193. I'll close it in favor of this one.

ckanext/harvest/harvesters/ckanharvester.py
davidread: Yeah urllib2 can raise this. It is an annoyance of urllib2 - a bug I guess. See http://www.voidspace.org.uk/python/articles/urllib2.shtml#badstatusline-and-httpexception and https://stackoverflow.com/questions/666022/what-errors-exceptions-do-i-need-to-handle-with-urllib2-request-urlopen

davidread:
@amercader can you review/merge this then?

ckanext/harvest/harvesters/ckanharvester.py
metaodi: Okay, good to know. No wonder all the docs say you should use the Requests module... :wink:

raphaelstolt:
What's the status on this one? /cc @davidread @amercader

davidread:
I'm gonna merge now since @metaodi and @raphaelstolt are happy with it.

davidread:
This bugfix (handling invalid source urls better) is also cherry-picked to release-v2.0 as requested above and in #194

PR #200, opened 2015-11-25
Fix job stats key

davidread:
It is "errored" not "errors" - the keys are from report_status.

And lots of PEP8 in dictization.py

### DIFF
--- a/ckanext/harvest/logic/dictization.py
+++ b/ckanext/harvest/logic/dictization.py
@@ -1,9 +1,8 @@
 from sqlalchemy import distinct, func
-import ckan.logic as logic
 
-from ckan.model import Package,Group
-from ckanext.harvest.model import HarvestSource, HarvestJob, HarvestObject, \
-                                  HarvestGatherError, HarvestObjectError
+from ckan.model import Package, Group
+from ckanext.harvest.model import (HarvestSource, HarvestJob, HarvestObject,
+                                   HarvestGatherError, HarvestObjectError)
 
 
 def harvest_source_dictize(source, context):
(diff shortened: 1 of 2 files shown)


davidread:
@metaodi could you just check this PR is correct, since it relates to your PR #192 ?

metaodi:
@davidread Oh that means I accidentally changed the API without noticing it? Sorry about that.

I think it's odd that the key is different for the job statistics than [for the source statistics] (https://github.com/ckan/ckanext-harvest/blob/6c3efe53df5a0f731f346e4d4972fa4a0b673638/ckanext/harvest/logic/dictization.py#L118). Either key is okay for me, but I'd vote to make it consistent for jobs and source statistics. I'm not sure if this kind of breaking change is acceptable or not. If there are doubts, we can leave it as it is.

Thanks for fixing this!

davidread:
It's a good point, the stats are different in this one respect.

I don't think anything is done with the job stats, apart from displayed, so maybe it's ok to change. Do you think extensions would use them? Were you looking at the stats when you thought to do #192?

metaodi:
Well, we're building a WordPress extension as an "admin interface" for harvesters. So yes, we are consuming and using the API and therefore noticed these inconsistencies.
That said, I think it would be great to be consistent and I frankly think there are not a lot of extensions using the harvester API. At least not enough to not change this. What's your impression?

davidread:
Merging this, as it is really a one-line fix and @metaodi is satisfied.

I'll create a ticket to propose the change to make the statuses consistent.

PR #203, opened 2015-11-27
Take advantage of named exceptions.

davidread:


metaodi:
I like it, the more named exceptions, the better :+1:

davidread:
Over to you @amercader

PR #207, opened 2015-12-02
You can abort a job by specifying the ID of the job

davidread:
...rather than the source id/name. This is helpful since the "harvest run" command returns a list of still running job ids.


metaodi:
I like that, makes the handling a little easier.

PR #215, opened 2015-12-10
Remove genshi templates

### LINKED ISSUE #214: Remove old Genshi templates
These are most likely very broken

davidread:
Fixes #214


amercader:
@davidread It looks like the templates got moved like

ckanext/harvest/templates_new/index.html → 
ckanext/harvest/templates/templates_new/index.html


Is that what you intended? Or just


ckanext/harvest/templates_new/index.html → 
ckanext/harvest/templates/index.html

davidread:
Good spot! That's fixed and I've tested it actually renders now.

PR #213, opened 2015-12-09
[#212] Fixes #212 - auth for harvest_job_create was broken.

davidread:
Issue #212

PR #216, opened 2015-12-10
Fix tests for ckan 2.2

davidread:
Following on from https://github.com/ckan/ckanext-harvest/pull/213#discussion_r47223762

I did this fix on another branch before I saw that. That's also a way to do it, but it does break DRY as assert_in is available in the legacy tests in ckan 2.2, so neither solution is perfect. I can't see a lot of point trying to make the harvest tests compatible with ckan pre 2.2... But if you feel its better, I'll change to that way.


davidread:
Ok had some problem, so have switched to @amercader 's suggested way

Issue #220, opened 2016-01-20
Harvest sources are listed on organization detail page


metaodi:
Currently harvest sources are shown on a organization detail page. I'm not sure if this is a bug or a feature. I already created an issue on the ckan repo (https://github.com/ckan/ckan/issues/2803), but maybe here is the better place.

Generally it's not wrong that harvest sources are displayed, as they clearly belong to the organization. But users might be confused, as they don't know what a harvest source is and wouldn't expect it on the organization detail page

As @wardi pointed out, other extensions wrote [code to exclude their custom dataset types from the search results](https://github.com/ckan/ckanext-showcase/blob/da5e28b91c011aa8f5417082a6320d4d23021e78/ckanext/showcase/plugin.py#L244-L255). Should we add something similar to ckanext-harvest?

I'm happy to help with the PR.

Btw: to solve the problem for me, I added a piece of code to my custom extension, that always adds a filter for datasets with the type `dataset`. This works, but I'm not sure if everybody has the re-implement that.

LondonAppDev:
+1
I'm also looking into ways for hiding the harvesters from the organisation results.

davidread:
I'm in favour of your suggested change. It would be good to get a comment from @amercader about it though.

amercader:
Fixed in #221

PR #221, opened 2016-02-12
Exclude custom 'harvester' datatype from the search results

LondonAppDev:
Prevent the 'harvest' datasets being shown in the group and dataset search results.

This is the change discussed in this ticket: https://github.com/ckan/ckanext-harvest/issues/220

I applied the fix which @metaodi and @wardi suggested in the ticket: https://github.com/ckan/ckanext-showcase/blob/da5e28b91c011aa8f5417082a6320d4d23021e78/ckanext/showcase/plugin.py#L244-L255


amercader:
Thanks for the patch @LondonAppDev

PR #223, opened 2016-02-15
PEP8 / tidy

davidread:
Extracted from https://github.com/ckan/ckanext-harvest/pull/158


davidread:
Trivial PR for @amercader or @rossjones to merge if that's ok?

PR #224, opened 2016-03-02
Replace deprecated nav_named_link by nav_link

dlax:
Getting rid of warnings such as:
```
2016-03-02 10:41:16,658 WARNI [ckan.lib.maintain] Function nav_named_link() in module ckan.lib.helpers has been deprecated and will be removed in a later release of ckan. h.nav_named_link is deprecated please use h.nav_link
NOTE: you will need to pass the route_name as a named parameter
```

The deprecation is from CKAN 2.0 according to ckan/lib/helpers.py.


amercader:
Thanks for the patch @dlax

Issue #227, opened 2016-03-21
Action 'harvest_source_show_status' not found


smotornyuk:
This error appears sometimes after adding new source and restarting server.
[After this](https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/model/__init__.py#L521) call activated harvester's [after_view](https://github.com/datawagovau/ckanext-harvest/blob/master/ckanext/harvest/plugin.py#L89).  Action list at this moment already cached(before harvester plugin added) and we have KeyError as result.

amercader:
@smotornyuk can you send a PR with your fix?

PR #228, opened 2016-03-21
[#227] 'harvest_source_show_status' not found

smotornyuk:
Added try block and try to clear action's cache in case exception raised


amercader:
Thanks for the patch @smotornyuk

PR #229, opened 2016-03-23
Unicode encode error with scandinavian facet filters

Zharktas:
If there are scandinavian alhabets (ä,ö and å) in tags or some other facets, harvest will throw UnicodeEncodeError while filtering with those facet.

This small PR will fix this error, since the search url is given to harvest plugin.


brew:
Thanks!

PR #230, opened 2016-04-12
Logging module

polarp:
Implemented database log handler for the standard logging module.
Harvest logs are now exposed through the CKAN API.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -197,6 +197,11 @@ The following operations can be run from the command line using the
           WARNING: if using Redis, this command purges all data in the current
           Redis database
 
+      harvester clean_harvest_log
+        - Clean-up mechanism for the harvest log table.
+          You can configure the time frame through the configuration
+          parameter 'ckan.harvest.log_timeframe'. The default time frame is 30 days
+
       harvester [-j] [-o] [--segments={segments}] import [{source-id}]
         - perform the import stage with the last fetched objects, for a certain
           source or a single harvest object. Please note that no objects will
--- a/README.rst
+++ b/README.rst
@@ -108,6 +108,12 @@ If you are having troubles configuring harvest logger please refer to ``test-cor
 
    If no value is present the default is 30 days.
 
+9. Setup log level for the database logger::
+
+     ckan.harvest.log_level = info
+
+   If no log level is set the default is ``debug``.
+
 There are a number of configuration options available for the backends. These don't need to
 be modified at all if you are using the default Redis or RabbitMQ install (step 1). The list
 below shows the available options and their default values:
--- a/README.rst
+++ b/README.rst
@@ -58,7 +58,53 @@ running a version lower than 2.0.
 
      ckan.harvest.mq.type = redis
 
-7. Setup time frame(in days) for the clean-up mechanism with the following config parameter::
+7. If you want your ckan harvest logs to be exposed to the ckan API you need to properly 
+   configure the logger. The default configuration logs everything to the database with 
+   log level ``DEBUG``. If you want to modify the database logger configure the following 
+   parameter::
+
+    ``ckan.harvest.log_scope = 0``
+
+    * Log scope settings:
+
+        - ``-1`` Do not log to the database
+        - ``0``  Log everything - Default
+        - ``1``  model, logic.action, logic.validators, harvesters
+        - ``2``  model, logic.action, logic.validators
+        - ``3``  model, logic.action
+        - ``4``  logic.action
+        - ``5``  model
+        - ``6``  plugin
+        - ``7``  harvesters
+
+    Additionally you can configure the logger in the following way::
+
+       [loggers]
+       keys = ckan_harvester
+
+       [handlers]
+       keys = dblog
+
+       [formatters]
+       keys = dblog
+
+       [logger_ckan_harvester]
+       qualname = ckanext.harvest
+       handlers = dblog
+       level = DEBUG
+
+       [handler_dblog]
+       class = ckanext.harvest.log.DBLogHandler
+       args = ()
+       level = DEBUG
+       formatter = dblog
+
+       [formatter_dblog]
+       format = %(message)s
+
+ If you are having troubles configuring ckan logger please refer to ``test-core.ini`` 
+
+8. Setup time frame(in days) for the clean-up mechanism with the following config parameter::
 
      ckan.harvest.log_timeframe = 10
 
@@ -91,39 +137,6 @@ config option (or ``default``) will be used to namespace the relevant things:
 * On Redis, it will namespace the keys used, so only the relevant instance gets them, eg
   ``site1:harvest_job_id``,  ``site1:harvest_object__id:804f114a-8f68-4e7c-b124-3eb00f66202f``
 
-7. If you want your ckan harvest logs to be exposed to the ckan API you need to add the 
-   following configuration options in your ckan configuriation file::
-
-     [loggers]
-
-     keys = ckan_harvester
-
-     [handlers]
-
-     keys = dblog
-
-     [formatters]
-
-     keys = dblog
-
-     [logger_ckan_harvester]
-
-     qualname = ckanext.harvest
-     handlers = dblog
-     level = DEBUG
-
-     [handler_dblog]
-
-     class = ckanext.harvest.log.DBLogHandler
-     args = ()
-     level = DEBUG
-     formatter = dblog
-
-     [formatter_dblog]
-     
-     format = %(message)s
-
- If you are having troubles configuring ckan logger please refer to ``test-core.ini`` 
 
 Configuration
 =============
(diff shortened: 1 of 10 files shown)


amercader:
@polarp Thanks for submitting this patch. Can you give more context as to what the use case for this and why do you think it needs to be part of ckanext-harvest?

AFAICT this can all live into your own project extension, and I'm not sure it is useful enough for the majority of ckanext-harvest users.

ckanext/harvest/logic/action/get.py
```diff
@@ -310,6 +312,38 @@ def harvesters_info_show(context,data_dict):
...
+    :param offset: use with ``per_page`` default: 0
+    :param level: filter log entries by level(debug, info, warning, error, critical)
+    '''
+
+    check_access('harvest_log_list', context, data_dict)
+
+    model = context['model']
+    session = context['session']
+
+    per_page = data_dict.get('per_page', 100)
```
TkTech: This needs an enforced cap - ex: `min(int(data_dict.get('per_page', 100)), 100)`

ckanext/harvest/tests/test_action.py
```diff
@@ -253,12 +253,15 @@ def test_create(self):
...
             assert_equal(source_dict[key], result[key])
-
+            
         # Check that source was actually created
         source = harvest_model.HarvestSource.get(result['id'])
         assert_equal(source.url, source_dict['url'])
         assert_equal(source.type, source_dict['source_type'])
 
+        # New source is created so the harvest_log table should be populated
+        assert model.Session.query(harvest_model.HarvestLog).count() > 0
```
TkTech: Just testing existence tends to bite you later as your tests get more complex. Instead of checking the length of the results, look for and compare the actual `HarvestLog` result that you expected to be stored.

ckanext/harvest/log.py
```diff
@@ -0,0 +1,16 @@
...
+    def __init__(self):
+        super(DBLogHandler,self).__init__()
+
+    def emit(self, record):
+        try:
+            level = record.levelname
+            msg = self.format(record)
+            obj = HarvestLog(level=level, content=msg)
+            obj.save()
+        except:
```
TkTech: *Never* do a blank `except` - this catches everything, including things like KeyboardInterrupt (such as someone hitting ^C). If you must catch all exceptions, you `except Exception`. Better would be to catch the specific exceptions you expect to see here, or at least log the exception.

davidread:
I'm in favour of this feature. When a harvest fails it is good to know what it was doing at the time. And it's better to provide this to the user, rather than require the site's sysadmin trawl through the logs on their behalf. So I see this as potentially saving a lot of support requests.

It's because there are lots of hidden assumptions when you harvest. e.g.
* CKAN harvester assumes you find the API at the source.url+'/api' and just seeing a log line showing it is trying that will explain the error.
* CKAN harvester will not harvest records that have not changed since the last harvest. Also WAF harvester gets an index of records and then filters out some. Showing records that have not been updated can explain why the harvest appears to not be doing anything.

I have one concern about this code - wouldn't this store a huge amount of extra data? The logs are pages of text for every harvest, and a site could easily be harvesting dozens of sources daily.

Perhaps in the future we can expose this log info in the web interface too.

ckanext/harvest/logic/action/get.py
```diff
@@ -310,6 +312,38 @@ def harvesters_info_show(context,data_dict):
 
     return available_harvesters
 
+@side_effect_free
+def harvest_log_list(context,data_dict):
+    '''Returns a list of harvester log entries grouped by level.
```
amercader: AFAICT the output is a dict with keys being the log levels and values lists of messages

amercader:
@davidread I'm not saying this is not a potentially useful feature, but to me it seems like not specific to harvesting at all (you can store any logging in the db by configuring the handlers on the ini file). If you think it's convenient for harvester users we can definitely include it as part of the extension though.

@TkTech @polarp volume of data is definitely a concern. Harvester logs can be **very** verbose and there doesn't seem to be a cleanup mechanism. I think that at least we need a config option to define what level of messages do you want to store on the DB, regardless of the logging configuration set on the ini file.

Also I'm not sure that the current output is the most useful. AFAICT the output is a dict with keys being the log levels and values lists of messages. When looking at the logs you want to see all messages sequentially ordered by timestamp. Perhaps a single list of objects with `level`, `timestamp` and `message` would be easier to interpret.

ckanext/harvest/logic/action/get.py
polarp: @TkTech Thanks for the suggestion, I will fix it.

ckanext/harvest/log.py
polarp: @TkTech Thanks for the suggestions, I will fix it.

polarp:
@amercader This was requested by a customer so that's why I think it should be part of the extension.

@amercader @davidread I agree that the volume of data is a concern so I will add cleanup mechanism for the log table that can be configured by the user e.g. logs older than 1/2/3 months should be deleted.

The grouping of the output was also a customer requirement but since there is filter by level I will remove the grouping and change the output.

If you have any more suggestions please let me know.

ckanext/harvest/commands/harvester.py
```diff
@@ -87,7 +93,7 @@ class Harvester(CkanCommand):
 
       harvester job-all
         - create new harvest jobs for all active sources.
-
+https://www.facebook.com/
```
davidread: This is a bit random!

davidread:
You're adding a table - do we need a db migration for existing users?

ckanext/harvest/commands/harvester.py
polarp: Yep somehow I've inserted that by accident...I'll remove it. Thank you

amercader:
@polarp is this ready for final review? Including the migration for existing users?

polarp:
@amercader yes it is ready. All features from the comments are implemented including the migration for existing users

README.rst
```diff
@@ -58,6 +58,62 @@ running a version lower than 2.0.
 
      ckan.harvest.mq.type = redis
 
+7. If you want your ckan harvest logs to be exposed to the CKAN API you need to properly
```
amercader: Can you move all DB logger configuration steps (7 to 9) to a separate section after "Configuration", just to make clear that this is an optional step.

README.rst
```diff
@@ -58,6 +58,62 @@ running a version lower than 2.0.
...
 
+7. If you want your ckan harvest logs to be exposed to the CKAN API you need to properly 
+   configure the logger. The default configuration logs everything in the database with 
+   log level ``DEBUG``. If you want to modify the default logging mechanism set the following 
+   parameter in your configuration file::
+
+     ckan.harvest.log_scope = 0
+
+ * -1 - Do not log in the database
+ *  0 - Log everything - DEFAULT
```
amercader: Please change the default to -1 (feature disabled)

README.rst
```diff
@@ -58,6 +58,62 @@ running a version lower than 2.0.
...
+ *  1 - model, logic.action, logic.validators, harvesters
+ *  2 - model, logic.action, logic.validators
+ *  3 - model, logic.action
+ *  4 - logic.action
+ *  5 - model
+ *  6 - plugin
+ *  7 - harvesters
+
+Additionally you can configure the logger the following way::
```
amercader: This implies that the ini file logger configuration is another option, but in fact it's a requirement, so I think a wording like "Additionally, you need to configure the logger in the CKAN ini file in the following way" is more adequate.

ckanext/harvest/plugin.py
```diff
@@ -217,6 +218,9 @@ def configure(self, config):
 
         # Setup harvest model
         model_setup()
+        
+        # Configure logger
+        _configure_logger(config)
```
amercader: Rename to `_configure_db_logger` to make it more explicit

ckanext/harvest/plugin.py
```diff
@@ -463,3 +467,57 @@ def _delete_harvest_source_object(context, data_dict):
...
+    #  0 - log everything
+    #  1 - model, logic.action, logic.validators, harvesters
+    #  2 - model, logic.action, logic.validators
+    #  3 - model, logic.action
+    #  4 - logic.action
+    #  5 - model
+    #  6 - plugin
+    #  7 - harvesters
+    #
+    scope = p.toolkit.asint(config.get('ckan.harvest.log_scope', 0))
```
amercader: Please change the default to -1 (feature disabled)

amercader:
@polarp There are a couple of things to change regarding documentation and default values but other than that looks good to go.

@davidread feel free to have an extra look if you want to double check

amercader:
Also I although the DB logger is enabled on the tests I can't see any actual test for `harvest_log_list` working as expected. Can you add one?

README.rst
polarp: @amercader Actually it is an additional configuration option not a requirement so maybe it would be better if I remove it from the README and use it only for the tests?

polarp:
@amercader I'll fix the issues from the comments and let you know when I'm finished.

polarp:
@amercader Just to let you know that I updated the documentation and I wrote test for the db logger so everything should be good to go.

davidread:
I had another skim and it looks great to me. Thanks for addressing all these requests @polarp.

amercader:
Thanks for the feature @polarp!

polarp:
@amercader @davidread Np...I'm glad I was able to help!

PR #233, opened 2016-04-27
Create index to harvest_object

### LINKED ISSUE #232: harvest_job_list ridiculously slow
In our production we currently have about 550 harvest jobs and harvest object count is currently going at 280000. 

Since harvest_job_list fetches them all in https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/logic/action/get.py#L239
and dictize https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/logic/dictization.py#L37 filters them one by one, harvest_object table could use a index for harvest_job

for example CREATE INDEX harvest_job_id_index ON harvest_object(harvest_job_id)

In our production single dictize took 1.5 seconds before index and 1.5ms after index and this made possible to actually call harvest_job_list as nginx timeouted it at 2 minutes before index. After index it took only few seconds, which is reasonable as there are no public links for templates which use job_list.

Zharktas:
Fixes #232 

Should create index in new instances as well in old ones.

### DIFF
--- a/ckanext/harvest/model/__init__.py
+++ b/ckanext/harvest/model/__init__.py
@@ -8,6 +8,7 @@ from sqlalchemy import Table
 from sqlalchemy import Column
 from sqlalchemy import ForeignKey
 from sqlalchemy import types
+from sqlalchemy import Index
 from sqlalchemy.engine.reflection import Inspector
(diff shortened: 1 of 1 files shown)


davidread:
Works great

PR #239, opened 2016-05-13
Import dependencies in setup.py.

torfsen:
Currently, requirements are only listed in `pip-requirements.txt` and not in `setup.py`. Hence, after installing `ckanext-harvest` via `pip` one has to manually install the dependencies. This is described in the README, but people expect that `pip install ...` also installs dependencies. In addition, `setup.py` does list test requirements, but these differ from the ones listed in `dev-requirements.txt`.

This PR updates `setup.py` so that it automatically imports the list of packages required for installation and testing from `pip-requirements.txt` and `dev-requirements.txt`. This gives us DRY and `pip` users get what they expect.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,5 @@
 recursive-include ckanext/harvest/templates *
 recursive-include ckanext/harvest/fanstatic_library/styles *
 recursive-include ckanext/harvest/public *
+include pip-requirements.txt
+include dev-requirements.txt
--- a/README.rst
+++ b/README.rst
@@ -44,16 +44,12 @@ running a version lower than 2.0.
 
      (pyenv) $ pip install -e git+https://github.com/ckan/ckanext-harvest.git#egg=ckanext-harvest
 
-4. Install the python modules required by the extension::
-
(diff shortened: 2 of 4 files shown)


davidread:
There is some neat code here.

However perhaps we've not been clear enough, but there is a reason why we put the requirements in a file rather than setup.py. We have a policy for extensions of not putting dependencies in setup.py's install_requires because it is so inflexible. If an installer wants to use a different version to the recommended one in requirements.txt then you have to use a fork of the code or something else non-ideal. Python will not let you have one extension specifying redis==2.10.1 and another redis==2.10.2. Now you could argue every extension should specify a nice range of versions that it is tested against, but that is more work for an extension to keep up to date and test as soon as new ones appear. And it doesn't allow you to suggest a particular version for people to choose by default, that say the extension writer uses herself.

Am happy to have a wider debate on this, perhaps on the ckan ideas repo. And we should probably document our policy on this somewhere.

torfsen:
@davidread OK. In that case we should also remove the test requirements from `setup.py`. A documentation of that policy (and a link to it in `setup.py`) would be awesome!

Issue #240, opened 2016-05-19
Creation/import of groups from remote CKAN instance is borken


polarp:
Bug in harvester related to SQLAlchemy query when trying to check if a group exists

davidread:
Are you doing a broken test or PR? Or do you have a stacktrace?

polarp:
@davidread This is a PR. Here is the stacktrace:

`2016-05-19 12:59:31,404 DEBUG [ckanext.harvest.model] (ProgrammingError) can't adapt type 'dict' 'SELECT "group".id AS group_id, "group".name AS group_name, "group".title AS group_title, "group".type AS group_type, "group".description AS group_description, "group".image_url AS group_image_url, "group".created AS group_created, "group".is_organization AS group_is_organization, "group".approval_status AS group_approval_status, "group".state AS group_state, "group".revision_id AS group_revision_id \nFROM "group" \nWHERE "group".id = %(id_1)s \n LIMIT %(param_1)s' {'id_1': {u'display_name': u'\xd6ffentliche Verwaltung, Haushalt und Steuern', u'description': u'', u'image_display_url': u'', u'title': u'\xd6ffentliche Verwaltung, Haushalt und Steuern', u'id': u'51722fb1-dd7f-48dc-bceb-4738ad5cb169', u'name': u'verwaltung'}, 'param_1': 1}`

This is happening when the harvester is checking if the remote org is present in the local CKAN instance(the query was malformed) but after I fixed this multiple bugs occurred regarding the import of remote groups and organizations. As you can see [here](https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/harvesters/ckanharvester.py#L407) group_name is not the actual group name but a dict of the group obj.

Currently I'm working on fixing all issues that emerged from this bug.

amercader:
@polarp I guess your PR on core (ckan/ckan#3043) is with regards to these errors:

```
2016-05-23 10:51:47,159 DEBUG [ckanext.harvest.model] Action function organization_show did not call its auth function
```

I found those in the past, they generally mask a problem some where else, and I don't think your PR is necessary. This should be fixed in ckanext-harvest. Let me have a test to see if I can find what's going on.

amercader:
@polarp Please check if https://github.com/ckan/ckanext-harvest/pull/243 fixes your issues

polarp:
[...]

PR #242, opened 2016-05-20
Fixed creation/import of groups and organizations when harvesting from remote ckan instance

polarp:
This is a fix for #240

### DIFF
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -128,10 +128,10 @@ class CKANHarvester(HarvesterBase):
 
                 # Check if default groups exist
                 context = {'model': model, 'user': c.user}
-                for group_ in config_obj['default_groups']:
+                for group_name in config_obj['default_groups']:
                     try:
                         group = get_action('group_show')(
-                            context, {'id': group_['id']})
+                            context, {'id': group_name})
                     except NotFound, e:
                         raise ValueError('Default group not found')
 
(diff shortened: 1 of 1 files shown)


ckanext/harvest/harvesters/ckanharvester.py
```diff
@@ -123,10 +128,10 @@ def validate_config(self, config):
 
                 # Check if default groups exist
                 context = {'model': model, 'user': c.user}
-                for group_name in config_obj['default_groups']:
+                for group_ in config_obj['default_groups']:
```
amercader: This seems wrong. `default_groups` is (or should be) a list of of strings (names or ids). Why do you assume they are a dict?

ckanext/harvest/harvesters/ckanharvester.py
```diff
@@ -404,36 +409,30 @@ def import_stage(self, harvest_object):
                 # check if remote groups exist locally, otherwise remove
                 validated_groups = []
 
-                for group_name in package_dict['groups']:
+                for group_ in package_dict['groups']:
```
amercader: This makes sense, I assume it was a remnant of the old API version 1/2 based harvester where `groups` is a list of strings.

ckanext/harvest/harvesters/ckanharvester.py
polarp: @amercader You're right, I overlooked this. I'll revert that change.

polarp: @amercader Just to let you know that I reverted the change made in the default groups.

amercader:
@polarp Thanks for the patch. I added a couple of tests for the remote group creation in 16a6e9fbf6eb53, you can use them as guidance if you need to send further patches.

Issue #185, opened 2015-11-04
Email admins when harvest fails


davidread:
This appears to be implemented by the US team (props @yaditi):

https://github.com/ckan/ckanext-harvest/compare/stable...GSA:2.3-upgrade#diff-b14f3f9793a50861d8d323e1cf51123bR370

It would be great to have this in core for the benefit of everyone - what do people think?

davidread:
@amercader is encouraging, as long as it's optional.

Consider per-source config to send emails always, when broken or never. Consider tying this into the 'follow' stuff already in CKAN.

earthquakesan:
@davidread: the government portal we are working on (Leipzig city) is dealing with quite a number of harvesting job. Some of them are long running and manually check when they are over is not really cool.

Imo, community will benefit from the feature as emailing harvest_job creator when it's done.

davidread:
Great to hear it @earthquakesan . My concern about merging this work is that it will generate emails that whilst helpful for many, other recipients will want to opt-out and they won't be able to. data.gov doesn't use a web-form so all their users are focussed on harvesting, but that's not true for most other sites.

So I imagine we need to extend the concept of a user 'following' a dataset to 'followers' of a harvester. And editors/admins of an organization would be set to follow a harvester when it is created, and the emails have a footer with a link to 'unfollow' and stop the emails. It would be great if anyone has a chance to take this on.

earthquakesan:
@davidread so, as I see it:
1. Make "Follow" button for harvest jobs which adds users (admins) to the user_following_harvest_job table
2. Refactor [https://github.com/GSA/ckanext-harvest/blob/809d6888aa1e7baa614ed16611df64b2f6ae604f/ckanext/harvest/logic/action/update.py](commit you mentioned originally) into a separate notification python module
3. Change the logic [https://github.com/GSA/ckanext-harvest/blob/809d6888aa1e7baa614ed16611df64b2f6ae604f/ckanext/harvest/logic/action/update.py#L459](for selecting emails) to get them from the user_following_harvest_job table

Did I understand you correctly?

davidread:
@earthquakesan yes I think you've got a good plan here.

> 1. Make "Follow" button for harvest jobs which adds users (admins) to the user_following_harvest_job table

[...]

PR #244, opened 2016-06-03
Send error mail to admin when harvesting fails

stefina:
This fixes https://github.com/ckan/ckanext-harvest/issues/185

I integrated a part of the https://github.com/GSA/ckanext-harvest solution into the ckanext-harvest-module.

To use it, configure the needed configuration values: 
**ckan.harvest.status_mail.errored = true**
email_to = %a comma seperated list of people to receive the error notification-mails%
error_email_from = %email-adress of sender%
ckan.site_title = %name of your site, so receivers know who you are%

How it works:
After being done with harvesting, the harvest_object-table will be checked for errors and the e-mail is generated accordingly. While the GSA-module sends an email after every harvesting-job about all states (added, updated, deleted) this only takes 'errored' into consideration.

### DIFF
--- a/ckanext/harvest/logic/action/update.py
+++ b/ckanext/harvest/logic/action/update.py
@@ -24,12 +24,14 @@ from ckan.logic import NotFound, check_access
 from ckanext.harvest.plugin import DATASET_TYPE_NAME
 from ckanext.harvest.queue import get_gather_publisher, resubmit_jobs
 
-from ckanext.harvest.model import HarvestSource, HarvestJob, HarvestObject
+from ckanext.harvest.model import HarvestSource, HarvestJob, HarvestObject, HarvestObjectError, HarvestGatherError
 from ckanext.harvest.logic import HarvestJobExists
 from ckanext.harvest.logic.dictization import harvest_job_dictize
 
 from ckanext.harvest.logic.action.get import (
     harvest_source_show, harvest_job_list, _get_sources_for_user)
+
+import ckan.lib.mailer as mailer
 
 log = logging.getLogger(__name__)
 
@@ -549,6 +551,9 @@ def harvest_jobs_run(context, data_dict):
                     # status
                     get_action('harvest_source_reindex')(
                         context, {'id': job_obj.source.id})
+
+                    if toolkit.asbool(config.get('ckan.harvest.status_mail.errored')) and has_job_errors(context, job_obj):
+                        send_error_mail(context, job_obj)
                 else:
                     log.debug('Ongoing job:%s source:%s',
                               job['id'], job['source_id'])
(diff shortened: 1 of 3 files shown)


amercader:
@stefina many thanks for submitting this!

I have some comments on the actual implementation but before going into them I just want to clarify a couple of things.

* You say that `this only takes 'errored' into consideration`, but unless I'm missing something this will send an email every time a harvest job finishes (ie `send_error_mail` is always called)

* You need to explicitly define the recipients of the error emails on your ini file, which seems very close to say that only sysadmins will receive this notifications, because you are not filtering which organizations people will receive emails fo (also `error_to` is meant to be used by sysadmins only to get site exception emails). This might still be a useful use case but it's not what was discussed in #244, if I understood correctly. There the approach was that publishers got notified of their own sources failures.

@davidread @metaodi @earthquakesan you on commented on #244 so you might have a view on this.

metaodi:
For our use case, running a harvester is pretty much sysadmin work, so using error_to is sensible. More sophisticated ways to determine recipients of such emails could be implemented later if someone wants to.

stefina:
@amercader thank you very much for your feedback!

As for your assumptions:
1. You are right, the mail should only be sent out, if actually an error occures during harvesting. As of right now it does always check if something went wrong but sends an e-mail either way. That I will change. (Only for the case that it is wanted to receive this mail in any case, wether it failed or succeeded, this would make sense.)

2. Yes, exactly, as @metaodi said, our use case is to send it only to the sysadmins, that is why I didn't include the functionality to send it to the contact-points.

davidread:
It'd be good to get this big of code merged in and at some point we can extend it to mailing non-sysadmins at some point.

Would you mind changing these raw SQL statements to use sqlalchemy ORM, to match the ckan style? i.e. model.Session.query() instead of using model.Session.execute(sql). I think it is more readable too as the joins are already defined in the mapper.

stefina:
@davidread Thank you for your feedback! Definitely think that makes sense, so I will change the queries to ckan-style.

ckanext/harvest/logic/action/update.py
```diff
@@ -482,6 +484,9 @@ def harvest_jobs_run(context, data_dict):
                     # status
                     get_action('harvest_source_reindex')(
                         context, {'id': job_obj.source.id})
+
+                    if toolkit.asbool(config.get('ckan.harvest.status_mail.errored')):
```
amercader: As discussed, we should only set the email if the job errored.

ckanext/harvest/logic/action/update.py
```diff
@@ -491,6 +496,88 @@ def harvest_jobs_run(context, data_dict):
...
+
+    q = model.Session.execute(sql, {'source_id': job_obj.source.id})
+
+    for row in q:
+        harvest_name = str(row['name'])
+
+    ckan_site_url = config.get('ckan.site_url')
+    job_url = ckan_site_url + '/harvest/' + harvest_name + '/job/' + job_obj.id
+
+    msg = 'This is a failure-notification of the latest harvest job (' + job_url + ') set-up in ' + ckan_site_url + '.\n\n'
```
amercader: All user facing strings should be translatable, you need to use something like:

```python
msg = p.toolkit._('This is a failure-notification of the latest harvest job ({job_url}) set-up in {ckan_site_url}').format(job_url, ckan_site_url)
msg += '\n\n'
```

ckanext/harvest/logic/action/update.py
```diff
@@ -491,6 +496,88 @@ def harvest_jobs_run(context, data_dict):
...
+
+    sql = '''select g.title as org, s.title as job_title from member m
```
amercader: Depending on your authorisation setup not all harvest sources might belong to an organization (for instance if `ckan.auth.create_unowned_dataset = true`). You should consider that when building the output.

ckanext/harvest/logic/action/update.py
```diff
@@ -491,6 +496,88 @@ def harvest_jobs_run(context, data_dict):
 
     return []  # merely for backwards compatibility
 
+def send_error_mail(context, job_obj):
+
+    sql = 'select name from package where id = :source_id;'
```
amercader: What @davidread said about ORM statements

ckanext/harvest/logic/action/update.py
```diff
@@ -491,6 +496,88 @@ def harvest_jobs_run(context, data_dict):
...
+    sql = '''select g.title as org, s.title as job_title from member m
+            join public.group g on m.group_id = g.id
+            join harvest_source s on s.id = m.table_id
+            where table_id = :source_id;'''
+
+    q = model.Session.execute(sql, {'source_id': job_obj.source.id})
+
+    for row in q:
+        msg += 'Organization: ' + str(row['org']) + '\n\n'
+        msg += 'Harvest Job Title: ' + str(row['job_title']) + '\n\n'
```
amercader: This is actually the Harvest Source Title

ckanext/harvest/logic/action/update.py
```diff
@@ -491,6 +496,88 @@ def harvest_jobs_run(context, data_dict):
...
+
+    msg += 'Date of Harvest: ' + str(job_obj.created) + ' GMT\n\n'
```
amercader: Harvest job `created` is created as an UTC date, better to use this than GMT to avoid confusion, even if they are nearly always the same.

ckanext/harvest/logic/action/update.py
```diff
@@ -491,6 +496,88 @@ def harvest_jobs_run(context, data_dict):
...
+    }
+
+    out['last_job'] = harvest_job_dictize(job_obj, context)
+
+    msg += 'Records in Error: ' + str(out['last_job']['stats'].get('errored', 0)) + '\n'
+
+    obj_error = ''
+    job_error = ''
+
+    sql = '''select hoe.message as msg from harvest_object ho
```
amercader: Some sources can have hundreds of remote documents and depending on the harvester you might get hundreds of errors so I'll suggest putting a hard limit of 20 or something reasonable or you might end up with a huge email.

ckanext/harvest/logic/action/update.py
```diff
@@ -491,6 +496,88 @@ def harvest_jobs_run(context, data_dict):
...
+        msg += 'Document Error\n' + obj_error + '\n\n'
+
+    if (job_error != ''):
+        msg += 'Job Errors\n' + job_error + '\n\n'
+
+    if obj_error or job_error:
+        msg += '\n--\nYou are receiving this email because you are currently set-up as Administrator for ' + config.get('ckan.site_title') + '. Please do not reply to this email as it was sent from a non-monitored address.'
+
+        # get recipients
+        email_recipients = config.get('email_to').split(',')
```
amercader: `email_to` is a paster setting for getting the stacktraces. Let's have a specific setting for this called `ckan.harvest.status_mail.email_to` instead, and you can check if this is set on `harvest_jobs_run` rather than having a separate `ckan.harvest.status_mail.errored`

amercader:
@stefina I added a few comments on the implementation please let me know if you have any doubts.

Remember also to document this functionality in the README

bonnland:
I have tried using this "harvest email notification" pull request, and I was getting this exception when making the call to mailer.mail_recepient():

2017-09-11 18:05:08,510 ERROR [ckanext.harvest.logic.action.update] No object (name: app_globals) has been registered for this thread

So I added this line just before the call to mailer.mail_recipient:  

pylons.app_globals._push_object(config['pylons.app_globals'])

Was this a terrible hack?   It seemed to fix the problem, i.e. multiple "paster harvester run" commands each generate an email notification if there are harvest errors.

ckanext/harvest/logic/action/update.py
rossjones: I think it would be cleaner to move 569-581 into a separate function (to check it a particular job errored) and include the output of that here, if `ckan.harvest.status_mail.errored` is True.

Other than that, I think it looks good.

@stefina I realise that this is an old PR, but it's really useful functionality .. do you think you'd be up for the small change suggested?

stefina: @rossjones Good point! I refactored it accordingly.

stefina:
I finally got around to work on this PR. Would appreciate your feedback or get this merged if it looks okay to you. As far as I know I changed my implementation according to your feedback from earlier.

So far, error-mails will only be sent to configured sys-admins, no further distinction of data-publishers or contact-points whatsoever.

rossjones:
Although this does only email system administrators, which is good for one use-case, I think it gives most of the necessary code to do the same for administrators of the harvester organization, which would be added in another PR.  

It really _should_ have a test, but I forgot to mention that in my feedback. :( If you don't have time to add a test @stefina let us know and I'll try and do one in a separate PR (as I can't push to your branch) sometime in the next couple of days.

ckanext/harvest/logic/action/update.py
```diff
@@ -559,6 +564,116 @@ def harvest_jobs_run(context, data_dict):
...
+    if (job_error != ''):
+        msg += toolkit._('Job Errors')
+        msg += '\n' + job_error + '\n\n'
+
+    if obj_error or job_error:
+        msg += '\n--\n'
+        msg += toolkit._('You are receiving this email because you are currently set-up as Administrator for {0}. Please do not reply to this email as it was sent from a non-monitored address.').format(config.get('ckan.site_title'))
+
+        # get recipients
+        email_recipients = config.get('email_to').split(',')
```
amercader: If the goal is to send it to all sysadmin users it is much safer to query them from the database as I suspect `email_to` is not always filled:

```python
sysadmins = model.Session.query(model.User).filter(model.User.sysadmin==True).all()

for user in sysadmins:
   
   # Use user.email


```

stefina:
@amercader thank you for those hints, this simplifies the code immensely!
Was not entirely sure if I still like the solution of getting the recipients from `email_to` better, as I assume it is there for a reason and I consider it a little easier to configure it there (and it is documented as of now; still need to change the readme) but I also don't know about the majority of CKAN-Setups so I changed it to use the DB-Admins

@rossjones I finally got to implement the tests. I am not entirely sure if the Exception-Handling with the MailerException is best practice:
https://github.com/stefina/ckanext-harvest/blob/00911a69249b85291f42da759911b789522ab46b/ckanext/harvest/logic/action/update.py#L559-L562
Neither am I sure that this is the right way to deal with that exception:
https://github.com/stefina/ckanext-harvest/blob/00911a69249b85291f42da759911b789522ab46b/ckanext/harvest/tests/test_action.py#L638-L646

Furthermore I can imagine that there might be a more straight-forward way to fake a HarvestGatherError than I did here:
https://github.com/stefina/ckanext-harvest/blob/00911a69249b85291f42da759911b789522ab46b/ckanext/harvest/tests/test_action.py#L630-L634
While writing this I am realizing that my test fails with prior CKAN-Versions, so I will have a look into that now.

Either way, I hope those changes get this PR a step closer to being merged :D

amercader:
@stefina can you extend your test class from `helpers.FunctionalTestBase` like this [one](https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/tests/test_controller.py#L24) to see if that helps with the failures?

stefina:
Thank you @amercader that did the trick!
BTW: Through mocking the `mailer.mail_recipient`-function I was able to remove the unneeded Boolean and Error-Handling in the calling function that I added just for the sake for testing.

amercader:
:dancer: :tada: great work @stefina !

This is great, thanks for your effort and patience with the review comments.

I agree that emailing all sysadmins might not be the best option but IMO is better than to rely on `email_to`, which is really a maintenance setting to get exception emails. Other options that could work better would be:

* Have a dedicated config option for emails that should be notified when a job fails (eg `ckan.harvest.error_emails.send_to`)
* Notify the administrator of the organization/publisher that the harvest source belongs to. This is probably the one that makes more sense, as are the ones most likely to be able to act on any potential error. In case you or someone else wants to pick this up in the future it's quite easy to get org admins rather than sysadmins:

```python

members = get_action('member_list')(context, {'id': source['organization']['id']})
for member in members:
    if member['capacity'] == 'admin':
        # add to recipients (we might need an extra call to user_show to get the email)

```

davidread:
Nice work @stefina !!

PR #252, opened 2016-06-14
Don't let the user specify mutually exclusive configuration options

jweyrich:
This change prevents the user from saving a harvest source that contains both `organizations_filter_include` and `organizations_filter_exclude` in its configuration options.


amercader:
Thanks @jweyrich

PR #254, opened 2016-06-24
Fixes default_groups

### LINKED ISSUE #253: CKAN Harvester Issue With Default Groups
I've been trying to set up a CKAN installation for use as a Caribbean open data portal for a few weeks now but I've been running into challenges when I try to set up the CKAN Harvester and import some data sets. 

My issue has to do with me trying to set the "default_groups" parameter.

If I set it to an empty set like so --> "default_groups" : []

There are no issues.

But if I try to include groups like so ---> "default_groups" : ["trinidad-and-tobago"]

I get this error :

---

The form contains invalid entries:

Config: Error parsing the configuration options: default_groups must be a list of group names/ids (i.e. strings)

---

I've done my best to be thorough here but please feel free to ask me any additional questions for clarification. I'd be very grateful for any insight/help with figuring this out.

I posted a variation of this as a question on StackExchange here in case it helps in providing some context : http://opendata.stackexchange.com/questions/9063/ckan-harvester-issues

nick22891:
Fixes #253

This fixes the two interconnected issues where perfectly normal "default_groups" parameters were failing - both at the validation stage and the import stage.

The fix at the validation stage simply involves changing the type from a string to a basestring as David Read suggested.

The fix at the import stage was because it couldn't find the self.default_group_dicts variable that was being referenced. I made it available and populated it from the default groups variable already defined on line 493.

It works great for me now and I've tested this fix on two separate servers. Cheers!

### DIFF
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -128,7 +128,7 @@ class CKANHarvester(HarvesterBase):
                     raise ValueError('default_groups must be a *list* of group'
                                      ' names/ids')
                 if config_obj['default_groups'] and \
-                        not isinstance(config_obj['default_groups'][0], str):
+                        not isinstance(config_obj['default_groups'][0], basestring):
                     raise ValueError('default_groups must be a list of group '
                                      'names/ids (i.e. strings)')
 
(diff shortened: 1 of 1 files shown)


davidread:
Thanks for the input on this. I've taken this and done a more efficient PR https://github.com/ckan/ckanext-harvest/pull/255 - hope you don't mind.

PR #255, opened 2016-06-27
[#253] Fix default_groups

davidread:
Saves the group dicts to the config object. Because saving them to the harvester object doesn't work in the real world (although it did in the tests).

Saving the group dicts up-front in a harvest is a lot more efficient than doing group_show for every dataset imported.

Also allow default_groups to be a list of *unicode* strings. I don't know how that would happen, but coping with this user's problem is no trouble.

### DIFF
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -128,20 +128,24 @@ class CKANHarvester(HarvesterBase):
                     raise ValueError('default_groups must be a *list* of group'
                                      ' names/ids')
                 if config_obj['default_groups'] and \
-                        not isinstance(config_obj['default_groups'][0], str):
+                        not isinstance(config_obj['default_groups'][0],
+                                       basestring):
                     raise ValueError('default_groups must be a list of group '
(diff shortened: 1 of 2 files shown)


davidread:
@nick22891 if you're able to try this variant of your fix, it would be much appreciated.

ckanext/harvest/harvesters/ckanharvester.py
```diff
@@ -128,20 +128,24 @@ def validate_config(self, config):
...
-                self.default_group_dicts = []
+                config_obj['default_group_dicts'] = []
                 for group_name_or_id in config_obj['default_groups']:
                     try:
                         group = get_action('group_show')(
                             context, {'id': group_name_or_id})
-                        self.default_group_dicts.append(group)
+                        # save the dict to the config object, as we'll need it
+                        # in the import_stage of every dataset
+                        config_obj['default_group_dicts'].append(group)
```
amercader: We just need `{"id": "xxx"}` but it's not a big deal

davidread: Ah, good point. I'll just leave it as it is though, if that's ok.

PR #258, opened 2016-07-20
[#257] Purge only our own Redis data.

### LINKED ISSUE #257: Allow sharing of Redis database
CKAN core is currently moving towards Redis as a hard dependency, for example for its new background job queue (ckan/ckan#3165). This means that CKAN and ckanext-harvest should be able to share a Redis database.

ckanext-harvest already [prefixes its Redis keys appropriately](https://github.com/ckan/ckanext-harvest/blob/7f506913f8e78988899af9c1dd518dc76e2c3e62/ckanext/harvest/queue.py#L70), but Redis queue purging [currently empties the whole database](https://github.com/ckan/ckanext-harvest/blob/7f506913f8e78988899af9c1dd518dc76e2c3e62/ckanext/harvest/queue.py#L101). This must be fixed so that only the ckanext-harvest keys of the current CKAN instance are removed. That can be done using Redis' support for LUA scripts, see for example the [queue cleaning in RQ](https://github.com/nvie/rq/blob/766bb600060c85e8a39579541798930bbb508ec8/rq/queue.py#L82).

torfsen:
Fixes #257.

Previously purging the queue on the Redis backend would clear the whole database, making it hard to share the same database with other parts of CKAN. With this commit, only the keys that belong to ckanext-harvest and the current CKAN instance are purged.

### DIFF
--- a/ckanext/harvest/queue.py
+++ b/ckanext/harvest/queue.py
@@ -76,19 +76,17 @@ def get_fetch_queue_name():
     return 'ckan.harvest.{0}.fetch'.format(config.get('ckan.site_id',
                                                       'default'))
 
-
 def get_gather_routing_key():
-    return '{0}:harvest_job_id'.format(config.get('ckan.site_id',
-                                                  'default'))
+    return 'ckanext-harvest:{0}:harvest_job_id'.format(
+            config.get('ckan.site_id', 'default'))
 
 
 def get_fetch_routing_key():
-    return '{0}:harvest_object_id'.format(config.get('ckan.site_id',
-                                                     'default'))
+    return 'ckanext-harvest:{0}:harvest_object_id'.format(
+            config.get('ckan.site_id', 'default'))
 
 
 def purge_queues():
-
     backend = config.get('ckan.harvest.mq.type', MQ_TYPE)
     connection = get_connection()
     if backend in ('amqp', 'ampq'):
@@ -97,10 +95,12 @@ def purge_queues():
         log.info('AMQP queue purged: %s', get_gather_queue_name())
         channel.queue_purge(queue=get_fetch_queue_name())
         log.info('AMQP queue purged: %s', get_fetch_queue_name())
-        return
-    if backend == 'redis':
-        connection.flushdb()
-        log.info('Redis database flushed')
+    elif backend == 'redis':
+        get_gather_consumer().queue_purge()
+        log.info('Redis gather queue purged')
+        get_fetch_consumer().queue_purge()
+        log.info('Redis fetch queue purged')
+
 
 def resubmit_jobs():
     '''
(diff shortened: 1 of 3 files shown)


ckanext/harvest/tests/test_queue.py
```diff
@@ -93,12 +96,10 @@ def setup_class(cls):
...
         ### make sure queues/exchanges are created first and are empty
-        consumer = queue.get_consumer('ckan.harvest.test.gather',
-                                      queue.get_gather_routing_key())
-        consumer_fetch = queue.get_consumer('ckan.harvest.test.fetch',
-                                            queue.get_fetch_routing_key())
-        consumer.queue_purge(queue='ckan.harvest.test.gather')
-        consumer_fetch.queue_purge(queue='ckan.harvest.test.fetch')
+        consumer = queue.get_gather_consumer()
+        consumer_fetch = queue.get_fetch_consumer()
+        consumer.queue_purge(queue=queue.get_gather_queue_name())
```
davidread: The param is ignored now, isn't it?

ckanext/harvest/queue.py
```diff
@@ -68,27 +68,25 @@ def get_connection_redis():
 
 
 def get_gather_queue_name():
-    return 'ckan.harvest.{0}.gather'.format(config.get('ckan.site_id',
-                                                       'default'))
+    return 'ckanext-harvest:{0}:gather'.format(
```
davidread: Changing the queue name seems like bad news for existing users that upgrade to this version of ckanext-harvest. Can't we keep it the same, for ease?

ckanext/harvest/queue.py
```diff
@@ -202,20 +202,44 @@ def consume(self, queue):
             key, body = self.redis.blpop(self.routing_key)
             self.redis.set(self.persistance_key(body),
                            str(datetime.datetime.now()))
+            print('SET IT!')
```
davidread: print left in by mistake?

ckanext/harvest/queue.py
```diff
@@ -202,20 +202,44 @@ def consume(self, queue):
...
         self.redis.delete(self.persistance_key(message))
 
-    def queue_purge(self, queue):
-        self.redis.flushdb()
+    def queue_purge(self, queue=None):
+        '''
+        Purge the consumer's queue.
+
+        The ``queue`` parameter exists only for compatibility and is
+        ignored.
```
davidread: Can't we get rid of the parameter now?

davidread:
Nice job doing all this. Just a few little things I've asked about.

ckanext/harvest/queue.py
torfsen: Fixed.

ckanext/harvest/queue.py
torfsen: Fixed.

ckanext/harvest/queue.py
torfsen: I don't think so. As far as I can tell, `RedisConsumer` tries to look like an AMQP channel, whose `queue_purge` method *does* take the parameter. To be honest, the separation between the AMQP and Redis back-ends isn't that clear in the code so I tried to make as little changes as necessary.

ckanext/harvest/tests/test_queue.py
torfsen: Only for Redis, AMQP still requires it as far as I understand.

torfsen:
Thanks for the review! I've fixed/clarified the points you mentioned.

The tests with CKAN 2.5 [still fail](https://travis-ci.org/ckan/ckanext-harvest/jobs/146121840), but that seems to be a problem with the test setup (failure is during the installation):
```
Downloading html5lib-0.999999999.tar.gz (245kB)
    100% |################################| 245kB 1.8MB/s 
    html5lib requires setuptools version 18.5 or above; please upgrade before installing (you have 12.0.5)
    Complete output from command python setup.py egg_info:
    html5lib requires setuptools version 18.5 or above; please upgrade before installing (you have 12.0.5)
```
The other tests (including CKAN `master`) [pass](https://travis-ci.org/ckan/ckanext-harvest/builds/146121832).

ckanext/harvest/queue.py
davidread: Ah, that's why - thanks

ckanext/harvest/tests/test_queue.py
davidread: Ignored only by redis backend.

davidread:
Thanks again for this - merging now.

Yep the failure with master is due to changes on master. If you have a chance to look at that as well, then that would be appreciated...

PR #263, opened 2016-08-25
Add new parameter return_last_job_status to harvest_source_list

metaodi:
In order to get a quick overview over successful/failed harvesters, a
call to `harvest_source_list` with `return_last_job_status=true` can be used
to get this information.
By default `return_last_job_status` is `False`, and hence the extra
resources to grab this information is not wasted for every call, but
only if the client requests is explicitly.
The original `status` field stays as-is, this PR introduces a new field
called `last_job_status` to return this information.

The returned information is gathered by a call to
`harvest_source_status_show`.

### DIFF
--- a/ckanext/harvest/logic/action/get.py
+++ b/ckanext/harvest/logic/action/get.py
@@ -130,7 +130,7 @@ def harvest_source_list(context, data_dict):
 
     sources = _get_sources_for_user(context, data_dict)
 
-    last_job_status = data_dict.get('return_last_job_status', False)
+    last_job_status = p.toolkit.asbool(data_dict.get('return_last_job_status', False))
 
     return [harvest_source_dictize(source, context, last_job_status) for source in sources]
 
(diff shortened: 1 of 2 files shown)


metaodi:
As soon as ckan/ckan#3181 is merged, the build using CKAN 2.5 should be fine.

davidread:
LGTM and useful

ckanext/harvest/logic/action/get.py
```diff
@@ -130,7 +130,9 @@ def harvest_source_list(context, data_dict):
 
     sources = _get_sources_for_user(context, data_dict)
 
-    return [harvest_source_dictize(source, context) for source in sources]
+    last_job_status = data_dict.get('return_last_job_status', False)
```
amercader: You might want to use `toolkit.asbool`, otherwise passing `return_last_job_status=False` as a GET parameter will become a string and be interpreted as `True`

metaodi: Good catch, I quickly added that :+1:

amercader:
thanks for the patch @metaodi

PR #264, opened 2016-08-29
Specify AGPL in setup.py in accordance with CKAN

### LINKED ISSUE #245: Readme says license is AGPL, setup.py says its MIT
Which one is it? This should be consolidated ;)

metaodi:
Fixes #245


davidread:
Since June 2015 the README.rst had AGPL - before it had nothing.

setup.py had it as MIT since the repo was created March 2011 and it was MIT probably because it was the license suggested by Cabinet Office at the time.

It's good to be more consistent, and AGPL is more of the CKAN spirit.

PR #265, opened 2016-09-18
Change 'redirect' calls to 'h.redirect_to'

argv0:
ckan.lib.base.redirect was removed in [1], guidance is to always use h.redirect_to instead.  Manifested as 503 errors in the harvest UI against an install of CKAN master branch. 

[1] https://github.com/ckan/ckan/commit/34f3f18e8893b5fc33047df94c88d4852c9a1f6b


amercader:
Thanks @argv0. Actually what we should always do from an extension is use `toolkit.redirect_to` (which uses `h.redirect_to`), but that can be fixed later on.

PR #268, opened 2016-10-28
Add clearsource history command

seitenbau-govdata:
Adds a `clearsource history` command / operation.

### DIFF
--- a/ckanext/harvest/logic/action/update.py
+++ b/ckanext/harvest/logic/action/update.py
@@ -237,7 +237,7 @@ def harvest_sources_job_history_clear(context, data_dict):
 
     job_history_clear_results = []
     # We assume that the maximum of 1000 (hard limit) rows should be enough
-    result = logic.get_action('package_search')(context, {'fq': '+type:"harvest"', 'rows': 1000})
+    result = logic.get_action('package_search')(context, {'fq': '+dataset_type:harvest', 'rows': 1000})
     harvest_packages = result['results']
     if harvest_packages:
         for data_dict in harvest_packages:
--- a/ckanext/harvest/logic/action/update.py
+++ b/ckanext/harvest/logic/action/update.py
@@ -241,8 +241,13 @@ def harvest_sources_job_history_clear(context, data_dict):
     harvest_packages = result['results']
     if harvest_packages:
         for data_dict in harvest_packages:
-            clear_result = get_action('harvest_source_job_history_clear')(context, {'id': data_dict['id']})
-            job_history_clear_results.append(clear_result)
+            try:
+                clear_result = get_action('harvest_source_job_history_clear')(context, {'id': data_dict['id']})
+                job_history_clear_results.append(clear_result)
+            except NotFound:
+                # Ignoring not existent harvest sources because of a possibly corrupt search index
+                # Logging was already done in called function
+                pass
 
     return job_history_clear_results
 
--- a/ckanext/harvest/logic/action/update.py
+++ b/ckanext/harvest/logic/action/update.py
@@ -225,6 +225,68 @@ def harvest_source_clear(context, data_dict):
     return {'id': harvest_source_id}
 
 
+def harvest_sources_job_history_clear(context, data_dict):
+    '''
+    Clears the history for all active harvest sources. All jobs and objects related to a harvest source will
+    be cleared, but keeps the source itself.
+    This is useful to clean history of long running harvest sources to start again fresh.
+    The datasets imported from the harvest source will NOT be deleted!!!
+
+    '''
+    check_access('harvest_sources_clear', context, data_dict)
+
+    job_history_clear_results = []
+    # We assume that the maximum of 1000 (hard limit) rows should be enough
+    result = logic.get_action('package_search')(context, {'fq': '+type:"harvest"', 'rows': 1000})
+    harvest_packages = result['results']
+    if harvest_packages:
+        for data_dict in harvest_packages:
+            clear_result = get_action('harvest_source_job_history_clear')(context, {'id': data_dict['id']})
+            job_history_clear_results.append(clear_result)
+
+    return job_history_clear_results
+
+
+def harvest_source_job_history_clear(context, data_dict):
+    '''
+    Clears all jobs and objects related to a harvest source, but keeps the source itself.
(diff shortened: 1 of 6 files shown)


ckanext/harvest/logic/action/update.py
```diff
@@ -225,6 +225,68 @@ def harvest_source_clear(context, data_dict):
...
+    be cleared, but keeps the source itself.
+    This is useful to clean history of long running harvest sources to start again fresh.
+    The datasets imported from the harvest source will NOT be deleted!!!
+
+    '''
+    check_access('harvest_sources_clear', context, data_dict)
+
+    job_history_clear_results = []
+    # We assume that the maximum of 1000 (hard limit) rows should be enough
+    result = logic.get_action('package_search')(context, {'fq': '+type:"harvest"', 'rows': 1000})
```
amercader: The Solr index field that holds the dataset type is called `dataset_type`, so this query does no return anything. You should change it to `{'fq': '+dataset_type:harvest', 'rows': 1000})`. Note that if you want to wrap the value in double quotes we need to this change in plugin.py:

```diff
diff --git a/ckanext/harvest/plugin.py b/ckanext/harvest/plugin.py
index 55af1c7..fa57103 100644
--- a/ckanext/harvest/plugin.py
+++ b/ckanext/harvest/plugin.py
@@ -92,7 +92,7 @@ class Harvest(p.SingletonPlugin, DefaultDatasetForm, DefaultTranslation):
         '''Prevents the harvesters being shown in dataset search results.'''
 
         fq = search_params.get('fq', '')
-        if 'dataset_type:harvest' not in fq:
+        if 'dataset_type:harvest' not in fq and 'dataset_type:"harvest"' not in fq:
             fq = u"{0} -dataset_type:harvest".format(search_params.get('fq', ''))
             search_params.update({'fq': fq})
 
```
Which is an excellent idea anyway.

amercader:
@seitenbau-govdata if you merge master to solve the conflicts and add a small test we can merge this one

ckanext/harvest/logic/action/update.py
```diff
@@ -225,6 +225,68 @@ def harvest_source_clear(context, data_dict):
...
+    be cleared, but keeps the source itself.
+    This is useful to clean history of long running harvest sources to start again fresh.
+    The datasets imported from the harvest source will NOT be deleted!!!
+
+    '''
+    check_access('harvest_sources_clear', context, data_dict)
+
+    job_history_clear_results = []
+    # We assume that the maximum of 1000 (hard limit) rows should be enough
+    result = logic.get_action('package_search')(context, {'fq': '+dataset_type:harvest', 'rows': 1000})
```
davidread: Why add the limit? e.g. in data.gov.uk we have approximately 20,000 harvested datasets, so this would be no good.

amercader: This is searching for harvest sources, not harvested datasets. It could potentially be an issue on larger instances but I think it's good enough for a first version.

davidread: ah, good point @amercader. We have 400 of those. Still, why not remove the limit?

seitenbau-govdata:
Does anybody have an idea why with CKAN 2.2 in the test "test_harvest_sources_job_history_clear" is only one harvest source in the result instead of two? This is the reason why the travis ci build is failing on CKAN 2.2.
For me it seems like the solr index isn't updated with the second harvest source yet while the action "harvest_sources_job_history_clear" reads the existent harvest sources with "package_search".

ckanext/harvest/logic/action/update.py
davidread: Fair enough then - thanks for explaining. Perhaps add a note in the documentation about the limit, just in case?

davidread:
@seitenbau-govdata I did a little tidy of the use of test fixtures, including the one you added test_harvest_sources_job_history_clear:

https://github.com/ckan/ckanext-harvest/commit/117f0374a757b02a35107d85b2e5a0c39a570992 

Perhaps you'd like to add that to your branch? Otherwise I can add it as a separate PR.

I created a PR anyway https://github.com/ckan/ckanext-harvest/pull/271 to see if it happens to fix the ckan 2.2 problem.

BTW I had a quick look at that the ckan 2.2 test failure, but I couldn't see why it occurs. If @amercader can't see why then I guess someone needs to set-up ckan 2.2 and debug with that?

davidread:
Interesting - I ran the tests again and they pass! Both for you and on the alternative PR with my test tidies.

Perhaps let me know if you're happy with the test tidies and we can choose which branch for @amercader to review.

seitenbau-govdata:
@davidread Really strange that the tests are now successful. Your test tidies looking good for me. Thanks!
I have added your commit to our branch, too. So it would be nice when @amercader could use our branch for the review.

davidread:
Great, lgtm

amercader:
Nice work @seitenbau-govdata, thanks for the PR and thanks @davidread for your help reviewing

davidread:
I was thinking this is a useful command to clear out unused info about old harvests - it builds up. But a problem has been pointed out - on the next harvest it'll create duplicate datasets, because there are no HarvestObjects any more to link up a source dataset's GUID to the local dataset. (This might not be the case for all harvesters - some may use another method to identify datasets.)

(I guess for the usage I'm thinking of, we should keep the most recent HarvestObject for every GUID, to prevent the duplicates.)

So is this a bug, or maybe @seitenbau-govdata's harvester doesn't have this problem, or is there some other intended purpose for this command? It's just that leaving the harvest source implies you want to reharvest, so why keep the datasets if you're going to get duplicates?

At the least I've added a PR to warn about the duplicates problem:

seitenbau-govdata:
Hi @davidread Sorry for the late response. Your assumption is right. We just need the paster command to get rid of the old harvest jobs which reserve a big amount of data (in this case just waste) in the database with the time.

Maybe you can you show me the piece of code which causes the problems when the harvest object is no more present?

We don't have the problem with duplicates after re-harvesting when all harvest jobs are deleted. But we only use the dcat harvester from ckanext-dcat as base. The only thing is that in the stats after the first re-harvesting all updated datasets are shown as "created".

As suggested from you I also think it's much better to preserve the current harvest jobs instead of delete them all.

PR #270, opened 2016-11-15
Improve resolving local groups

seitenbau-govdata:
Improve resolving local groups by searching for a group additionally by name.

### DIFF
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -428,8 +428,20 @@ class CKANHarvester(HarvesterBase):
 
                 for group_ in package_dict['groups']:
                     try:
-                        data_dict = {'id': group_['id']}
-                        group = get_action('group_show')(base_context.copy(), data_dict)
+                        try:
+                            if 'id' in group_:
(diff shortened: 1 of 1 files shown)


amercader:
Just so I understand the logic behind the changes, you seem to take into account a scenario where the group id is present but not found, but the group name is found right? Also is there a scenario where the data_dict does not have either an `id` or `name`?

seitenbau-govdata:
Yes, right. Normally the id of the groups between two CKAN instances are not equal, except the remote groups were created. If there was still a group in the remote CKAN with the same name but with a different id I would assume that the group remains in the harvested dataset, because the name is also an unique attribute of the group.

And in the actual code the harvesting aborts when a harvested dataset contains a group dict without an 'id'. I don't know how realistic is the scenario that a group dict has no 'id' and 'name', but with the change the harvesting will continue and don't abort.

seitenbau-govdata:
@amercader I just wondering that the pull request isn't merged yet. Is there something still unclear, missing or to preparing/improving?

amercader:
Sorry @seitenbau-govdata, this slipped under my radar

seitenbau-govdata:
No problem. Thanks a lot, @amercader!

PR #278, opened 2017-01-13
Fix setup.py to match it's current paster template

mikkonieminen:
Fixes plugin's setup.py namespace_packages to match the updated setup.py paster template. Fixes the issue mentioned in e.g. here: https://github.com/ckan/ckan/issues/2893

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ setup(
 	url='http://ckan.org/wiki/Extensions',
 	license='AGPL',
 	packages=find_packages(exclude=['ez_setup', 'examples', 'tests']),
-	namespace_packages=['ckanext', 'ckanext.harvest'],
+	namespace_packages=['ckanext'],
 	include_package_data=True,
 	zip_safe=False,
 	install_requires=[


torfsen:
@mikkonieminen Note that the problem in ckan/ckan#2893 was that extensions did not declare `ckanext` as a namespace package. That some extensions additionally declare their `ckanext.<name>` package to be a namespace package is usually not required but shouldn't cause any problems.

(That also applies to your other PRs, datagovuk/ckanext-report#13, ckan/ckanext-qa#41, ckan/ckanext-archiver#32)

mikkonieminen:
@torfsen We are having a strange issue when installing the extensions with pip and Ansible but, when the `ckanext.<name>` namespace package is removed the issue goes away. Installation with `python setup.py develop` however works either way.

torfsen:
@mikkonieminen That's interesting. I've encountered a similar issue: you cannot mix editable and non-editable installations for sub-packages of the same namespace-package using `pip`. See pypa/pip#3 for details. As a result we're installing all our CKAN extensions in editable mode (`pip install -e ...`).

Since `ckanext.<name>` usually isn't required to be a namespace package (in contrast to `ckanext`) I see no problem in removing it (as your PR does) if that fixes your problems.

amercader:
Yes, this caused some obscure problems in the past, thanks @mikkonieminen

PR #283, opened 2017-02-21
#282 Return an empty list when no CKAN datasets are gathered

### LINKED ISSUE #282: Erroneous "Gather stage failed" error logged when there are no CKAN datasets to fetch
There appears to be mismatch in `gather_callback()`'s expectations of return values and actual return values for `gather_stage()` when no new/updated datasets are found to harvest. 

`gather_callback()` appears to want a [list of dataset IDs](https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/queue.py#L310-L314) even if the list is empty (Null Object Pattern) while `gather_stage()` returns `None` ([1](https://github.com/ckan/ckanext-harvest/blob/57173ff9fe6cf5a909d156d1c02c41442f9398b8/ckanext/harvest/harvesters/ckanharvester.py#L234), [2](https://github.com/ckan/ckanext-harvest/blob/57173ff9fe6cf5a909d156d1c02c41442f9398b8/ckanext/harvest/harvesters/ckanharvester.py#L253)) if no new or updated datasets are found.

MrkGrgsn:
Fixes #282

Doesn't include new tests; happy to look at that if wanted.

### DIFF
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -231,7 +231,7 @@ class CKANHarvester(HarvesterBase):
                 log.info('No datasets have been updated on the remote '
                          'CKAN instance since the last harvest job %s',
                          last_time)
-                return None
+                return []
 
         # Fall-back option - request all the datasets from the remote CKAN
         if get_all_packages:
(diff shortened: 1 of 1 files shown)


davidread:
Yes this makes sense.

The docs in interface.py are pretty clear that None is not allowed to be returned by gather_stage. There are number of places where None is returned in ckanext-spatial and ckanext-dcat harvesters, but it occurs after a gather error has been logged, so it's no big deal doing a log.error.

There is one more exception though - https://github.com/ckan/ckanext-dcat/blob/e1245cdb5e29194787c33430db637be74e5bc2e1/ckanext/dcat/harvesters/base.py#L210 which we should probably fix too, by probably adding a gather error and then returning [].

davidread:
Thanks @MrkGrgsn

PR #288, opened 2017-05-02
Minor test improvements 

davidread:
* More explicit checks of the exception thrown when checking harvest config
* The default_groups test was checking default_tags, so I fixed that (L274)


amercader:
Thanks @davidread

PR #295, opened 2017-08-07
Change urllib2 to urllib3

thenets:
I followed the security steps recommend on "urllib3" and "certifi" documentation.
This new version had all security fixes for TLS 1.2 and use "certifi" plugin certificates.

And I also create a script to generate PEM file with custom certificates. You can check at:
https://github.com/thenets/ckanext-harvest/tree/dadosgovbr/ca-custom

I need help to test all those changes.


thenets:
The test returns: `OK (SKIP=3)`
I used this container for CKAN:
https://github.com/thenets/Easy-CKAN/blob/dev/src/docker/2.6/ckan/Dockerfile

wardi:
@tino097 will take a look

pip-requirements.txt
```diff
@@ -1,2 +1,5 @@
 pika==0.9.8
 redis==2.10.1
+urllib3
+certifi
+requests[security]
```
tino097: overall it is looking good, it will be good to add package version to reduce chance for conflicts with other extensions

thenets: Ok. I'll do it. Can I set the current latest version?

tino097: latest stable i guess will be best

thenets: Ok! I'll commit it today.

thenets:
Python community from IRC recommended me to use the "requests" lib instead of the "urllib3". I'll migrate all modification to this lib and notify you when I finish.

amercader:
+1 just use requests and it will save you a lot of troube.

I personally like to keep support for older versions for as long as possible in extensions. If you feel strongly about it you can remove tests for 2.2 and 2.3, but 2.4 and 2.5 should still be supported.

thenets:
@amercader I'll add support to requests. I learned more about it just last week.
About support, I tried to install and doesn't work. Travis CI got success but I can't install into my VM. I don't know what should I do.

amercader:
@thenets just work on the latest version that you use for general development, and let Travis handle the different versions.

thenets:
@amercader Hm, ok. I'll add those versions again. I'm working on requests now.

brucebolt:
I'm wondering if there is any plan to get this PR included into ckanext-harvest?

amercader:
@brucebolt not as is, but it would be great to migrate to requests. It should be trivial, just replacing [this urllib2](https://github.com/ckan/ckanext-harvest/blob/f715aa5e1271bbc758de0ee72cc636577351a5aa/ckanext/harvest/harvesters/ckanharvester.py#L37) usage with `requests.get`() and adapt the exception codes

Issue #112, opened 2014-11-24
Catch more errors to avoid limbo states


davidread:
I've come a cross a bunch of ways to end up with the harvester in a 'limbo' state, so I'd like to add various exception handling and state management to avoid this better.

'limbo' is where the job state is 'running' yet something has gone wrong at it has stopped. The main problem is that you can't restart the job while it is supposedly still running.

e.g. you are writing a harvester and it has an uncaught exception in the gather, fetch or import_stage, then you enter limbo. This should be sorted better by queue.py, rather than rely on harvesters not raising exceptions.

e.g. you install a new harvester into a CKAN, restart CKAN but don't restart the queue. Running the job you get "No harvester could be found for source type x" and the job is in limbo.

torfsen:
Exceptions from a plugin's `gather_stage`, `fetch_stage`, and `import_stage` should be caught and the appropriate action should be taken. For example, there is already error handling code in [`queue.fetch_and_import_stages`](https://github.com/ckan/ckanext-harvest/blob/13d5395bf01e80e3883e7b08b01b1cdcf5734720/ckanext/harvest/queue.py#L391), but it's only active when the plugin reports the error, not if it throws an exception. Therefore, when my plugin raised an exception from its `import_stage` it crashed the fetch consumer and left the job in limbo.

What's the current state regarding this issue?

torfsen:
Also, the README for `gather_stage` says

> to abort the harvest, create a HarvestGatherError and raise an exception. Any created HarvestObjects will be deleted.

What it does not say is that this will crash the gather consumer because it re-raises the exception after doing the documented cleanup.

I don't see why so much of the error handling is left to the plugin implementation and why some of it is handled via return codes (i.e. returning `False` from `fetch_stage` and `import_stage`). From a plugin developer's point of view, here's how I'd like things to work:

If any of `gather_stage`, `fetch_stage` or `import_stage` raise an exception then the corresponding consumer catches it, logs it, makes sure the job and its objects are in a proper state, and then continues working. If no exception is raised then that already indicates success, so there's no need to pass an error flag via the return value.

If I'd need to create customized `HarvestGatherError` instances or some other fancy error handling then I can still do that in my plugin -- but I won't have to, and things will still work (or better, fail) correctly.

davidread:
I agree, the harvesting framework should be defensive against harvesting code that raises exceptions.

[...]

PR #296, opened 2017-10-25
Abort stuck harvest jobs through UI

aykutsimseks:
Occasionally harvester will get stuck at `Running` state (limbo) if it fails without proper error handling  and once it does the only way to restore harvester is either abort the job from CLI or the `Clear` option from UI. 

This PR provides UI for aborting limbo jobs without requiring to log into the server or clear the source altogether.

Relates to: #112

### DIFF
--- a/ckanext/harvest/templates/source/admin.html
+++ b/ckanext/harvest/templates/source/admin.html
@@ -6,12 +6,6 @@
     {% if source.status and source.status.last_job %}
       {% snippet "snippets/job_details.html", job=source.status.last_job %}
       <div class="form-actions">
-        {% if source.status and source.status.last_job and (source.status.last_job.status == 'Running') %}
-        <a href="{{ h.url_for('harvest_job_abort', source=source.name, id=source.status.last_job.id) }}" class="btn" title="Stop this Job">
-          <i class="icon-ban-circle"></i>
-          {{ _('Stop') }}
-        </a>
-        {% endif %}
         <a href="{{ h.url_for(controller='ckanext.harvest.controllers.view:ViewController', action='show_last_job', source=source.name) }}" class="btn pull-right">
           <i class="icon-briefcase"></i>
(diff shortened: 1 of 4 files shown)


metaodi:
The travis build is currently broken, my PR to fix it is here: #297

metaodi:
Okay the build is now fixed.

@aykutsimseks please rebase to master to get the fixed build.

aykutsimseks:
@metaodi thanks for the fix. PR rebased and updated. Build is green.

ckanext/harvest/templates/source/admin.html
```diff
@@ -6,6 +6,12 @@
     {% if source.status and source.status.last_job %}
       {% snippet "snippets/job_details.html", job=source.status.last_job %}
       <div class="form-actions">
+        {% if source.status and source.status.last_job and (source.status.last_job.status == 'Running') %}
+        <a href="{{ h.url_for('harvest_job_abort', source=source.name, id=source.status.last_job.id) }}" class="btn" title="Stop this Job">
+          <i class="icon-ban-circle"></i>
+          {{ _('Stop') }}
+        </a>
+        {% endif %}
```
metaodi: Instead of putting this new button in the bottom part next of the "View full job report", I would rather put it next to the other actions in the upper part between the "Reharvest" and the "Clear" action.

aykutsimseks: 👍  My initial inclination was, all the above buttons are for actions related to the source and the below is for the last job.  I'll move it between Reharvest and Clear.

metaodi: I understand and that could make sense as well. But e.g. the reharvest depends on the last job as well, as you can't start a new one, if one is currently running.

metaodi:
@aykutsimseks thanks a lot!

PR #298, opened 2017-10-26
Use the package_search API for the count of datasets

metaodi:
Rather than relying on the 'current' flag on the HarvestObject it's more
accuarte to use the package_search API to display the current number of
datasets, as this is actually the number of datasets displayed in the
list.
If for whatever reason the 'current' flag is not up-to-date, this leads
to a mismatch between the displayed number of datasets and the acutal
number of datasets returned by CKAN.

### DIFF
--- a/ckanext/harvest/helpers.py
+++ b/ckanext/harvest/helpers.py
@@ -61,6 +61,17 @@ def package_list_for_source(source_id):
 
     return out
 
+def package_count_for_source(source_id):
+    '''
+    Returns the current package count for datasets associated with the given
+    source id
+    '''
+    fq = 'harvest_source_id:"{0}"'.format(source_id)
+    search_dict = {'fq': fq}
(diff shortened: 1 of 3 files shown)


metaodi:
@amercader what do you think? I'm unsure where/if the total_dataset calculations are used somewhere else. Alternatively to this I could create a new template helper function to [display the number on the harvest source page](https://github.com/ckan/ckanext-harvest/blob/57173ff9fe6cf5a909d156d1c02c41442f9398b8/ckanext/harvest/templates/source/read_base.html#L31) and leave the API untouched.

amercader:
That sounds good @metaodi. 
Can you explain the changes in the tests? Shouldn't `total_datasets` remain the same with the changes? (it's fine if not, just want to double check)

metaodi:
@amercader I now changed the template to use a template helper and keep the original `total_datasets` field as-is.

PR #304, opened 2017-11-07
#303: fix clean_tags with tags dict

### LINKED ISSUE #303: Failing to clean up tag vocabularies
The configuration
```
  {"clean_tags": true}
```
will break the harvester in case remote tags are imported as dict and not plain strings.

The error is:
```
2017-09-19 10:08:43,652 ERROR  [ckanext.harvest.harvesters.base] [base.py:371] -- ord() expected a character, but string of length 13 found
Traceback (most recent call last):
  File "/usr/lib/ckan/default/src/ckanext-harvest/ckanext/harvest/harvesters/base.py", line 291, in _create_or_update_package
    tags = [munge_tag(t) for t in tags if munge_tag(t) != '']
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/munge.py", line 108, in munge_tag
    tag = substitute_ascii_equivalents(tag)
  File "/usr/lib/ckan/default/src/ckan/ckan/lib/munge.py", line 98, in substitute_ascii_equivalents
    if ord(char) in char_mapping:
TypeError: ord() expected a character, but string of length 13 found
```

and it happens because the `munge` function is looping not on the [legacy string format](https://github.com/ckan/ckanext-harvest/blob/v1.1.0/ckanext/harvest/harvesters/base.py#L231-L235), but [on a dict](https://github.com/ckan/ckanext-harvest/blob/v1.1.0/ckanext/harvest/harvesters/base.py#L244-L245) like this:
```
{u'vocabulary_id': None, u'state': u'active', u'display_name': u'Zone a rischio naturale', u'id': u'073080c8-fef2-4743-9c9e-6216019f8b3d', u'name': u'Zone a rischio naturale'}
```
so the var `char` will be holding the `vocabulary_id` string.

(Initally I was thinking this issue was related to #154, but atm tags from vocabularies are properly imported.)

etj:
Fix #303

### DIFF
--- a/ckanext/harvest/harvesters/base.py
+++ b/ckanext/harvest/harvesters/base.py
@@ -288,9 +288,21 @@ class HarvesterBase(SingletonPlugin):
 
             if self.config and self.config.get('clean_tags', False):
                 tags = package_dict.get('tags', [])
-                tags = [munge_tag(t) for t in tags if munge_tag(t) != '']
-                tags = list(set(tags))
-                package_dict['tags'] = tags
+                if tags:
+                    if not isinstance(tags[0], dict):
+                        # REST format: 'tags' is a list of strings
+                        tags = [munge_tag(t) for t in tags if munge_tag(t) != '']
+                        tags = list(set(tags))
+                        package_dict['tags'] = tags
+                    else:
+                        # package_show form: 'tags' is a list of dicts
+                        clean_tags = []
+                        for t in tags:
+                            m = munge_tag(t.get('name'))
+                            if m != '':
+                                t['name'] = m
+                                clean_tags.append(t)
+                        package_dict['tags'] = clean_tags
 
             # Check if package exists
             try:
--- a/ckanext/harvest/harvesters/base.py
+++ b/ckanext/harvest/harvesters/base.py
@@ -288,21 +288,7 @@ class HarvesterBase(SingletonPlugin):
 
             if self.config and self.config.get('clean_tags', False):
                 tags = package_dict.get('tags', [])
-                if tags:
-                    if not isinstance(tags[0], dict):
-                        # REST format: 'tags' is a list of strings
-                        tags = [munge_tag(t) for t in tags if munge_tag(t) != '']
-                        tags = list(set(tags))
-                        package_dict['tags'] = tags
-                    else:
-                        # package_show form: 'tags' is a list of dicts
-                        clean_tags = []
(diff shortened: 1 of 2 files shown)


ckanext/harvest/harvesters/base.py
```diff
@@ -288,9 +288,21 @@ def _create_or_update_package(self, package_dict, harvest_object,
 
             if self.config and self.config.get('clean_tags', False):
                 tags = package_dict.get('tags', [])
-                tags = [munge_tag(t) for t in tags if munge_tag(t) != '']
-                tags = list(set(tags))
-                package_dict['tags'] = tags
+                if tags:
+                    if not isinstance(tags[0], dict):
```
metaodi: Please do not check the type like this (and the `if tags:` is not necessary). Python uses the [EAFP style](http://python.net/~goodger/projects/pycon/2007/idiomatic/handout.html#eafp-vs-lbyl), so you should rather wrap this in a try/except and handle the exception, something like this:

```python
try:
    # assume it's in the package_show form
   tags = [munge_tag(t['name']) for t in tags if munge_tag(t['name']) != '']
   ...
except TypeError: # a TypeError is raised if `t` above is a string
   # REST format: 'tags' is a list of strings
   tags = [munge_tag(t) for t in tags if munge_tag(t) != '']
   ...

File: ckanext/harvest/harvesters/base.py [hunk 2]
```diff
@@ -288,9 +288,21 @@ def _create_or_update_package(self, package_dict, harvest_object,
...
+                        # REST format: 'tags' is a list of strings
+                        tags = [munge_tag(t) for t in tags if munge_tag(t) != '']
+                        tags = list(set(tags))
+                        package_dict['tags'] = tags
+                    else:
+                        # package_show form: 'tags' is a list of dicts
+                        clean_tags = []
+                        for t in tags:
```
metaodi: Use a list comprehension just like above, no need for a `for` loop here.

etj: Uhm ok, but we're updating a field in a dict, guess I'll need something more complex

ckanext/harvest/harvesters/base.py
etj: ok

etj:
WIll also move the logic for cleaning the tags in its own method for better testing it

metaodi:
Thank you @etj

PR #305, opened 2017-11-10
Move _last_error_free_job from CKANHarvester to HarvesterBase

knudmoeller:
* Since CKANHarvester._last_error_free_job is such a useful method for any kind of harvester, this PR moves it to HarvesterBase.
* Method signature is changed from _last_error_free_job to last_error_free_job to make it part of the API of HarvesterBase (rather then an internal method).
* All occurrences of _last_error_free_job have been adjusted.

### DIFF
--- a/ckanext/harvest/harvesters/base.py
+++ b/ckanext/harvest/harvesters/base.py
@@ -2,7 +2,9 @@ import logging
 import re
 import uuid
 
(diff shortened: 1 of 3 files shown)


metaodi:
Thanks @knudmoeller for this PR!

PR #311, opened 2018-02-05
Remove returning stored site user

### LINKED ISSUE #151: DetachedInstanceError when harvesting CKAN
When harvesting a [CKAN instance](http://open-data.yes-we-ckan.org/) (source install v2.5.0a as of July 2015) into another CKAN instance (datacats using latest master, v2.5.0a) with a custom ckanext-scheming dataset schema, I'm running into DetachedInstanceError indicating that the db conection via SQLAlchemy changes sessions.

```
DetachedInstanceError('Instance <User at 0x7fdf11ae5e50> is not bound to a Session; attribute refresh operation cannot proceed',)
```

My config:
```
{
 "api_version": 1,
 "user": "harvest",
 "remote_groups": "create",
 "remote_orgs": "create",
 "override_extras": true,
 "read_only": true,
 "force_all": true
}
```
I've created a sysadmin user called "harvest", same error as with my own sysadmin account.

I haven't seen a DetachedInstanceError since mentioned in the dev meeting archives a year ago. 
Has anyone seen this error lately, or is harvesting only broken on my end?

Zharktas:
Site user is lost when validation fail during package_create.

This might fix #151.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,3 +15,7 @@ services:
 install:
     - bash bin/travis-build.bash
 script: sh bin/travis-run.sh
+
+# the new trusty images of Travis cause build errors with psycopg2, see https://github.com/travis-ci/travis-ci/issues/8897
+dist: trusty
+group: deprecated-2017Q4
--- a/ckanext/harvest/harvesters/base.py
+++ b/ckanext/harvest/harvesters/base.py
@@ -169,9 +169,6 @@ class HarvesterBase(SingletonPlugin):
            ckanext.harvest.user_name = harvest
 
         '''
-        if self._user_name:
-            return self._user_name
-
         config_user_name = config.get('ckanext.harvest.user_name')
         if config_user_name:
             self._user_name = config_user_name


metaodi:
To fix the build add the following lines to `.travis.yml`:

```
# the new trusty images of Travis cause build errors with psycopg2, see https://github.com/travis-ci/travis-ci/issues/8897
dist: trusty
group: deprecated-2017Q4
```

metaodi:
@Zharktas what do you mean with "site user is lost"? Lost how? The method only returns the username and not a user object, and the username won't change. Or am I missing something?

Zharktas:
If the site user has been previously fetched, https://github.com/vrk-kpa/ckanext-harvest/blob/f5df983542781b69728e1f871f4e7fd5baa86bea/ckanext/harvest/harvesters/base.py#L192 this never executes as the. 

And as the validatation error happens here https://github.com/vrk-kpa/ckanext-harvest/blob/f5df983542781b69728e1f871f4e7fd5baa86bea/ckanext/harvest/harvesters/base.py#L356, the SQLalchemy detaches itself.

metaodi:
Okay just to get it right, if the user has been previously fetched (i.e. `self._user_name` is set) then the [following line](https://github.com/vrk-kpa/ckanext-harvest/blob/f5df983542781b69728e1f871f4e7fd5baa86bea/ckanext/harvest/harvesters/base.py#L192) is not executed:

```python
self._site_user = p.toolkit.get_action('get_site_user')(context, {})
```

So far, everything is clear.

What I don't understand is: why is it a problem if that line is not executed? Do we need `self._site_user` somewhere? In most cases `_get_user_name()` will just return `harvest` (btw: the above line is also not executed if there is a `harvest` user on the system).
My guess by looking at the source of [get_site_user()](https://github.com/ckan/ckan/blob/c63675252734b8d54bf8ebfc6fd759cb2dd070dc/ckan/logic/action/get.py#L2407-L2435) is, that the `model.Session.flush()` plays a role, but I'm not sure.

I don't think, that it's a good idea to call `get_site_user()` for every dataset in the harvester, so to save the username in a variable seems like a good idea.

Zharktas:
I'll test tomorrow if reorganizing the code will help the issue. Either validate the dataset before any SQL magic happens or catch validation errors properly.

Zharktas:
Closed in favor of #312

PR #312, opened 2018-02-08
Validate dataset before create or update

### LINKED ISSUE #151: DetachedInstanceError when harvesting CKAN

Zharktas:
Attempt number 2 to fix #151.

Dataset is now validated before it is given to the baseharvester, if there is validation error it is raised again to store it to the harvester log.

### DIFF
--- a/ckanext/harvest/harvesters/ckanharvester.py
+++ b/ckanext/harvest/harvesters/ckanharvester.py
@@ -9,6 +9,7 @@ from ckan.logic import ValidationError, NotFound, get_action
 from ckan.lib.helpers import json
 from ckan.lib.munge import munge_name
 from ckan.plugins import toolkit
+import ckan.lib.plugins as lib_plugins
 
 from ckanext.harvest.model import HarvestObject
 
@@ -527,6 +528,17 @@ class CKANHarvester(HarvesterBase):
                 # and saving it will cause an IntegrityError with the foreign
                 # key.
                 resource.pop('revision_id', None)
+
+
+            package_plugin = lib_plugins.lookup_package_plugin('dataset')
+            schema = package_plugin.create_package_schema()
+
+            context = {}
+            data, errors = lib_plugins.plugin_validate(
+                package_plugin, context, package_dict, schema, 'package_create')
+
+            if errors:
+                raise ValidationError(errors)
 
             result = self._create_or_update_package(
                 package_dict, harvest_object, package_dict_form='package_show')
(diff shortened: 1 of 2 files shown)


ckanext/harvest/harvesters/ckanharvester.py
```diff
@@ -528,6 +529,33 @@ def get_extra(key, package_dict):
...
+            if validate_packages:
+                if 'type' not in package_dict:
+                    package_plugin = lib_plugins.lookup_package_plugin()
+                    try:
+                        # use first type as default if user didn't provide type
+                        package_type = package_plugin.package_types()[0]
+                    except (AttributeError, IndexError):
+                        package_type = 'dataset'
+                        # in case a 'dataset' plugin was registered w/o fallback
+                        package_plugin = lib_plugins.lookup_package_plugin(package_type)
```
metaodi: Shouldn't that line be one level higher? Otherwise `package_plugin` is not set if there was no exception on line 540.

Zharktas:
I'll commit some tests and check that comment once I've cured of this flu of mine, so probably in couple of days or so.

ckanext/harvest/harvesters/ckanharvester.py
Zharktas: package_plugin is set on line 537 and most of that code is actually copy paste from here https://github.com/ckan/ckan/blob/master/ckan/logic/action/create.py#L136

metaodi: Ah perfect, thanks for the explanation!

frafra:
Is it something that can be rebased/reworked?

PR #313, opened 2018-02-13
FontAwesome 4.x/CKAN 2.7.x compatibility

metaodi:
CKAN 2.7.x upgrades to FontAwesome 4.x, so the CSS classes need to be
updated in order to work. But because this extension is used in older
CKAN versions as well, we simply add the new classes, while keeping the
old ones for backwards compatibility.

The new names are according to the upgrade guide: https://github.com/FortAwesome/Font-Awesome/wiki/Upgrading-from-3.2.1-to-4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,3 +15,7 @@ services:
 install:
     - bash bin/travis-build.bash
 script: sh bin/travis-run.sh
+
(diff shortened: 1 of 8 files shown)


amercader:
Thanks @metaodi

PR #320, opened 2018-04-29
Update redirect_to in View controller

Engerrs:
When I clicked on Clear/Stop/Reharvest buttons, while my CKAN setup was on /data, got wrong redicrect with two /data. I updated the redirect by adding one missing param.


amercader:
Does this affect non-root instances? ie does it still work if ckan is not set up in `/data`?

PR #323, opened 2018-05-16
CKAN Harvester option to include/exclude groups

nmchau87:
Similar to organization include/exclude option


rossjones:
I know this seems a little superfluous, but do you think you'd have a little time to add tests in the same sort of vein as https://github.com/ckan/ckanext-harvest/blob/f715aa5e1271bbc758de0ee72cc636577351a5aa/ckanext/harvest/tests/harvesters/test_ckanharvester.py#L145-L163 ?

PR #325, opened 2018-05-31
Update 'harvester sources by organization' page to work with latest CKAN

brucebolt:
I have been trying to use the display at '/organization/harvest/{id}' but noticed the code appears to not be compatible with the latest version of CKAN (as far as I can see, this page isn't actually displayed anywhere on CKAN by default).

This PR fixes those issues and allows the page to display correctly.

### DIFF
--- a/ckanext/harvest/controllers/organization.py
+++ b/ckanext/harvest/controllers/organization.py
@@ -2,7 +2,7 @@ import logging
 from urllib import urlencode
 
 from ckan import plugins as p
-from ckan.lib.base import c, model, request, render, h, g
+from ckan.lib.base import c, model, request, render, h
 from ckan.lib.base import abort
 import ckan.lib.maintain as maintain
 import ckan.lib.search as search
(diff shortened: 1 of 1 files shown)


rossjones:
Hi @brucebolt do you think you might have enough time to add a quick test (perhaps to https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/tests/test_controller.py )? 

I realise there isn't one currently for the OrganizationController there, but it'd be ace if we could add one so we catch this sort of things in fugure.

amercader:
The fact that is not linked from the UI anywhere or that `group_show` was used seems to suggest that this endpoint hasn't been used in a long time.

Does your site use it @brucebolt? If so we'll need to do a bit more work so the code works on all supported CKAN version (eg with your changes the facets won't work, modern CKAN versions use `h.facets()`.
My vote goes for deleting the controller though.

brucebolt:
We had planned to use this controller, but are not using it anymore, so this PR can be closed

amercader:
Great, I'll delete it then, thanks @brucebolt

PR #327, opened 2018-06-02
Add config option for dataset name append type

seitenbau-govdata:
Adds the possibility to set the append_type parameter used in _gen_new_name globally for all harvesters within the CKAN config. So the default type can be changed easily by setting it in the config without the need changing the source code. The default value is still the same and the possibility to overwrite the parameter on function call still exists, too.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -166,6 +166,22 @@ e.g. Fetch all logs with log level INFO:
     }
 
 
+Dataset name generation configuration (optional)
+================================================
+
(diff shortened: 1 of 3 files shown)


amercader:
@seitenbau-govdata thanks for your PR. So as I understand it the `append_type` parameter was already there but not being exposed higher up in the call stack, so `number-sequence` was always used.  This looks good to me, I'll let @metaodi have another look

PR #328, opened 2018-06-08
Fix print statements to be Py3 friendly

rossjones:
Fixes the print statements where they were doung to use print as a
function, and also to switch to .format() rather than %.

Also contains some flake8 whitespace changes as I suspect making this
PEP8ish will take several PRs.

### DIFF
--- a/ckanext/harvest/commands/harvester.py
+++ b/ckanext/harvest/commands/harvester.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 import sys
 from pprint import pprint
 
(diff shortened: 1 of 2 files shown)


metaodi:
Should be add something like flake8 to the travis build to actually enforce compliance? I never ran it against this codebase, but if the goal is PEP8, we should have a tool to check it.

rossjones:
Yes probably, however I think it'll be a huge PR so it might be best to pep8 piecemeal until it's pretty much there, otherwise nobody'll be able to merge anything :) 

The tests fails look like a network fail and possibly a test ordering issue.  Is it possible to restart the failed jobs?

PR #329, opened 2018-06-14
Send harvest-error-mails to organization-admins

stefina:
This adds the organization-admins to the error-mail-recipients if a harvest-job fails.

Now that the code is way simpler than it originally was, it is not such a big step to send the error-mails to the organization-admins as well.

If the harvest-source belongs to an organization we will query their users who have the admin-role and add them to the recipients, if they have an email configured.


amercader:
Thanks @stefina!

Issue #331, opened 2018-07-12
Support Redis password


gavinpc-mindgrub:
Thanks for creating and maintaining this extension.

We are using `ckanext-harvest` in a Heroku environment, where the Redis add-on has a URL that includes a u/p, like `redis://x:yz@host:port`.

It appears that Harvest cannot be used when Redis requires a password.

Would the following patch be an appropriate way to use the `ckan.harvest.mq.password` setting that is used for the `amqp` connection?

```
--- a/ckanext/harvest/queue.py
+++ b/ckanext/harvest/queue.py
@@ -64,6 +64,7 @@ def get_connection_redis():
     import redis
     return redis.StrictRedis(host=config.get('ckan.harvest.mq.hostname', HOSTNAME),
                           port=int(config.get('ckan.harvest.mq.port', REDIS_PORT)),
+                          password=config.get('ckan.harvest.mq.password', None),
                           db=int(config.get('ckan.harvest.mq.redis_db', REDIS_DB)))
```

Perhaps related is ckan/ckan#3165, which adds a `CKAN_REDIS_URL` option to core CKAN (as noted in #257).  When present, that setting might provide a better default now for the Redis connection.

amercader:
Fixed in #332

PR #332, opened 2018-07-12
Use Redis password from configuration when present.

gavinpc-mindgrub:
This PR is in reference to #331.

I have tested it in configurations with and without passwords, but I am not familiar enough with the build environment to add unit tests easily.


amercader:
Thanks @gavinpc-mindgrub, this looks really useful

PR #335, opened 2018-08-30
Ensures the AND operator for fq in solr

dennisrpnw:
Without the leading '+', the page of a harvest source shows all datasets available in ckan, not only the ones that belong to the source itself.
package_search() of the ckan core adds '+capacity:public ' in front of the harvest_source_id and without the + it then becomes an OR and all datasets get returned.


amercader:
I can't reproduce it (only datasets belonging to the source are shown to me) but it might depend on the Solr version and sounds sensible, so merging. Thanks @dennisrpnw

PR #336, opened 2018-10-19
Replace organization_read with organization.read

frafra:
The patch fixes this warning:

```
INFO  [ckan.lib.helpers] Route name "organization_read" is deprecated and will be removed.
Please update calls to use "organization.read" instead
```

### DIFF
--- a/ckanext/harvest/templates/snippets/source_item.html
+++ b/ckanext/harvest/templates/snippets/source_item.html
@@ -42,7 +42,7 @@ Example:
         {{ _('Datasets') }}: {{ source.status.total_datasets }}
       {% endif %}
       {% if not within_organization and source.organization %}
-      &mdash; {{ _('Organization') }}: {{ h.link_to(source.organization.title or source.organization.name, h.url_for('organization_read', id=source.organization.name)) }}</a>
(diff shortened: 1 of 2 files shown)


wardi:
IIUC for extensions designed to work with multiple CKAN versions we want to keep the old names.

frafra:
Oh, OK. I thought about it after seeing Travis testing the extensions for old versions of CKAN.

frafra:
Shoudn't be possible to have an if/else or try/except clause in this case, to avoid these messages?

metaodi:
The old names should be supported on all 2.x releases (maybe even on CKAN 3.0 to allow a smooth transition). The message is a simple deprecation warning, shouldn't hurt anyone.
But yes at some point in the future most extensions will have a BC break and no longer work for older versions of CKAN.

Theoretically you could use a if/else block and change the route name based on the version of CKAN. This sound very much like a feature request for the [ckantoolkit](https://github.com/ckan/ckantoolkit), a library build to ensure cross-version support for CKAN.

amercader:
Thanks @frafra. I think we are going to keep the old names until we can find an elegant cross-version solution like the one @metaodi mentions.

PR #338, opened 2018-10-25
A harvester may not have an info method

brucebolt:
This fixes the error `AttributeError: 'HarvesterBase' object has no attribute 'info'` which prevents new harvest sources from being created in some circumstances.

It also continues enforcing the validation, since a class without a info method still cannot be used as a valid harvest source type.

### DIFF
--- a/ckanext/harvest/logic/validators.py
+++ b/ckanext/harvest/logic/validators.py
@@ -113,7 +113,10 @@ def harvest_source_type_exists(value, context):
     # Get all the registered harvester types
     available_types = []
     for harvester in PluginImplementations(IHarvester):
-        info = harvester.info()
+        try:
+            info = harvester.info()
+        except AttributeError:
+            continue
         if not info or 'name' not in info:
             log.error('Harvester %s does not provide the harvester name in '
                       'the info response' % harvester)


amercader:
Thanks @brucebolt. We probably should raise an exception if the info method is not present as it's expected in some parts of the code like [here](https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/logic/action/get.py#L186). What was the situation in which you found a harvester without this method?

brucebolt:
Thanks for the quick response @amercader.

The case was that my tests for creating a harvest source in our plugin were failing because `HarvesterBase` in `ckanext-harvest` had no `info` method.  I think this is probably because `PluginImplementations(IHarvester)` is returning this base class, which appears to intentionally have no info method.

I don't think raising an exception would be the right thing here, as the base classes of harvester intentionally have no `info` method so they cannot be set up as an actual harvester.

metaodi:
@brucebolt I don't understand your use case. If you want to create a new harvest source, you should extend `HarvesterBase` and implement the `info()` method. If this method is not available, it should fail. Maybe a custom exception, as @amercader suggested, would be even better. But at least the `AttributeError` should be kept there.

brucebolt:
@metaodi In our use case, we have extended `HarvesterBase` but are getting an error `AttributeError: 'HarvesterBase' object has no attribute 'info'` when I execute `get_action('harvest_source_create')(.....)`, regardless of the harvester type I specify.

I suspect it is because `for harvester in PluginImplementations(IHarvester)` is pulling out all classes that implement `IHarvester`.  `HarvesterBase` is one such class.

This validation code is run when any type of harvest source is created, therefore I cannot currently create any harvest source because `HarvesterBase` does not have an `info` attribute.

metaodi:
@brucebolt Ok, I see. Thanks for the clarification. I just don't understand, why we have never seen this error before, since this code is old and `HarvesterBase` never had an `info()` method. But anyway I can see how in this validator it makes sense to ignore these harvesters and simply continue with the others.

elliotcm:
@metaodi Thanks for the reply.  I'm also working on this issue.

Are you happy to merge in @brucebolt's changes here, or would you prefer an alternate approach to avoiding this issue?

elliotcm:
Hi @metaodi 

Thanks for approving this.  Are there any further steps we need to take to get it merged and released?

metaodi:
@elliotcm previously the Travis CI build was failing, I just restarted it, as it seemed a temporary thing. Now everything is fine and I'll merge this.

elliotcm:
Perfect, thanks :)

PR #341, opened 2018-10-31
Fix SSL problems for old versions of Python 2.7.x

andreagrandi:
Hi,

with this patch we solved the connection probelms with urls using SSL when using an old version of Python, such as:

```
/usr/lib/venv/local/lib/python2.7/site-packages/requests/packages/urllib3/util/ssl_.py:122: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#insecureplatformwarning.
```

### DIFF
--- a/ckanext/harvest/templates/source/job/read.html
+++ b/ckanext/harvest/templates/source/job/read.html
@@ -6,7 +6,7 @@
 <div class="module-content">
 
   <p class="pull-right">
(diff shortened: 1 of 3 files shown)


andreagrandi:
I just realised we didn't update tests and they are failing. I will fix them too, cheers.

PR #344, opened 2018-11-01
Fix SSL problems for old versions of Python 2.7.x

andreagrandi:
Hi,

with this patch we solved the connection probelms with urls using SSL when using an old version of Python, such as:

```
/usr/lib/venv/local/lib/python2.7/site-packages/requests/packages/urllib3/util/ssl_.py:122: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#insecureplatformwarning.
```

### DIFF
--- a/pip-requirements.txt
+++ b/pip-requirements.txt
@@ -1,2 +1,9 @@
 pika==0.9.8
 redis==2.10.1
+requests==2.20.0
+idna==2.7
+ndg-httpsclient==0.5.1
+pyasn1==0.4.4
+pyOpenSSL==18.0.0
+urllib3==1.23
+enum34==1.1.6
--- a/pip-requirements.txt
+++ b/pip-requirements.txt
@@ -1,9 +1,4 @@
 pika==0.9.8
 redis==2.10.1
 requests==2.20.0
-idna==2.7
-ndg-httpsclient==0.5.1
-pyasn1==0.4.4
 pyOpenSSL==18.0.0
-urllib3==1.23
-enum34==1.1.6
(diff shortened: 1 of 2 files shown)


pip-requirements.txt
```diff
@@ -1,2 +1,9 @@
 pika==0.9.8
 redis==2.10.1
+requests==2.20.0
+idna==2.7
+ndg-httpsclient==0.5.1
+pyasn1==0.4.4
+pyOpenSSL==18.0.0
+urllib3==1.23
+enum34==1.1.6
```
metaodi: are all of these required? I can't see anything in the code about `idna`, `ndg-httpsclient`, `pyasn1` and `enum34`

andreagrandi: `idna` for example is needed by `pyOpenSSL` https://github.com/urllib3/urllib3/blob/master/src/urllib3/contrib/pyopenssl.py#L11
the others I'm not sure about. I'm porting changes my team did in the last 2 years https://github.com/datagovuk/ckanext-harvest/commits/2.0 in a for we have and we would really love to merge these back upstream. Do you have any suggestion about how I could check if the other mentioned dependencies may not be needed? Thanks

andreagrandi: Ok, we did more investigations on this and these are our findings:

* `enum34==1.1.6` and `idna==2.7` are being installed by `pyOpenSSL==18.0.0` (which we need to fix this TLS problem)
* `urllib3==1.23` is installed by `requests`

This leaves us with `ndg-httpsclient==0.5.1` and `pyasn1==0.4.4` which we are still not sure about. I will update this PR in the mean time and try to remove those libs from our environment to check if any other dependency is touched.

Cheers

andreagrandi: All right, I've removed the dependencies that were not needed. How does it look now? Cheers

metaodi:
@andreagrandi Thank you very much!

PR #348, opened 2018-12-03
Add an 'owner_org' to the v3 package migration

brucebolt:
The auto-migration fails because the new package does not have an owner organisation ID assigned, so fails validation


ckanext/harvest/model/__init__.py
```diff
@@ -574,6 +574,7 @@ def gen_new_name(title):
             'source_type': source.type,
             'config': source.config,
             'frequency': source.frequency,
+            'owner_org': source.publisher_id,
```
amercader: I don't think the harvest source `publisher_id` field is used on a standard CKAN instance but if it's useful on instances with customized schemas I'm happy to keep it.

PR #354, opened 2019-03-06
Apply flake8

Zharktas:
This PR adds flake8 test to travis and applys around 600 fixes to conform with pep8


Zharktas:
This PR also fixes build against ckan master where assert_in was removed from test helpers. Build against different versions should work if ubuntu mirrors wouldn't fail.

ckanext/harvest/model/__init__.py
```diff
@@ -242,92 +248,93 @@ def define_harvester_tables():
...
-        Column('config', types.UnicodeText, default=u''),
-        Column('created', types.DateTime, default=datetime.datetime.utcnow),
-        Column('type',types.UnicodeText,nullable=False),
-        Column('active',types.Boolean,default=True),
-        Column('user_id', types.UnicodeText, default=u''),
-        Column('publisher_id', types.UnicodeText, default=u''),
-        Column('frequency', types.UnicodeText, default=u'MANUAL'),
-        Column('next_run', types.DateTime),
-    )
+                                 Column('id', types.UnicodeText, primary_key=True, default=make_uuid),
```
davidread: I'd be tempted to put a line break in after `Table(` and keep everything to the left side, minimizing wrapping, which suits editing on small screens, or with side-by-side windows.

ckanext/harvest/plugin.py
```diff
@@ -228,20 +218,29 @@ def before_map(self, map):
         # (ie they are the ones for a package type)
         controller = 'ckanext.harvest.controllers.view:ViewController'
 
-        map.connect('{0}_delete'.format(DATASET_TYPE_NAME), '/' + DATASET_TYPE_NAME + '/delete/:id',controller=controller, action='delete')
-        map.connect('{0}_refresh'.format(DATASET_TYPE_NAME), '/' + DATASET_TYPE_NAME + '/refresh/:id',controller=controller,
-                action='refresh')
-        map.connect('{0}_admin'.format(DATASET_TYPE_NAME), '/' + DATASET_TYPE_NAME + '/admin/:id', controller=controller, action='admin')
-        map.connect('{0}_about'.format(DATASET_TYPE_NAME), '/' + DATASET_TYPE_NAME + '/about/:id', controller=controller, action='about')
-        map.connect('{0}_clear'.format(DATASET_TYPE_NAME), '/' + DATASET_TYPE_NAME + '/clear/:id', controller=controller, action='clear')
+        map.connect('{0}_delete'.format(DATASET_TYPE_NAME), '/' + DATASET_TYPE_NAME + '/delete/:id',
```
davidread: go on, break after all the commas ;-)

.travis.yml
```diff
@@ -20,3 +20,22 @@ script: sh bin/travis-run.sh
...
+      install:
+        - bash bin/travis-build.bash
+        - pip install flake8==3.5.0
+        - pip install pycodestyle==2.3.0
+      script:
+        - flake8 --version
+        # stop the build if there are Python syntax errors or undefined names
+        - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics --exclude ckan
+        # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
+        - flake8 . --count --max-line-length=127 --statistics --exclude ckan
```
davidread: This is the key decision here - that you removed `--exit-zero` compared to CKAN core, so ckanext-harvest is being more hard-line about style than CKAN core.
 
I support this - there's a trend towards this in python these days (see popularity of the [black](https://github.com/ambv/black) and automating this aspect of code review, the idea that there is no shame when you're being kept in line by a bot, rather than a human) and we have this opportunity to start from zero errors, so why not. CKAN core should probably move towards this too.

davidread:
Couple of suggestions, but LGTM. It's @amercader's extension so decides.

metaodi:
I really like this so I'm merging this now. I have a couple of things to add/change, but will do this after merging.

PR #356, opened 2019-03-15
Fix harvest request exceptions

kentsanggds:
We are currently getting a error AttributeError: `SSLError` object has no attribute `code`. 
This changes the harvest exception handling so we can correctly log the error.


metaodi:
Ah I see there is a CI error because of an unrelated import of new_test, I'll take care of it.

kentsanggds:
@metaodi cheers!

kentsanggds:
Found a bug, will deploy a new fix with tests

kentsanggds:
Will spin up another PR with the fix and tests

PR #359, opened 2019-03-17
Code style changes to model and plugin

metaodi:
This is a follow-up PR to #354.


metaodi:
This PR is based on the review feedback on #354 by @davidread.

PR #363, opened 2019-05-03
Add try except around Redis set to handle corrupt redis data

kentsanggds:
## What

`try / except` added to catch any errors from corrupt Redis data in order to prevents bad data from stopping harvest processing and log the error for investigation, full stack trace output as part of log.

## Reference 

https://trello.com/c/MIA4o7ro/946-investigate-and-fix-corrupted-redis-data-issue


kentsanggds:
Created PR by accident

PR #381, opened 2019-11-13
get message from harvest_object_error-dict

stefina:
When activating this feature for a new client, I realised that the `harvest_object_error` structure is very different from what `harvest_gather_error` looks like.

An example of a `harvest_object_error`:

```
'object_errors': {u'02aa088c-5272-4b71-bdb1-e2d887e020f2': {'errors': [{'line': None,
                                                                         'message': u'Ftplib error,
                                                                         'type': u'Fetch'}],
                                                             'guid': u'Custom Harvester'},
```

If I am getting something wrong, I'd be happy for advice.

Irony is, that the original code is mine and I wonder, why this ever worked 😅

### DIFF
--- a/ckanext/harvest/logic/action/update.py
+++ b/ckanext/harvest/logic/action/update.py
@@ -610,8 +610,10 @@ def send_error_mail(context, source_id, status):
     obj_error = ''
     job_error = ''
 
-    for harvest_object_error in islice(report.get('object_errors'), 0, 20):
-        obj_error += harvest_object_error['message'] + '\n'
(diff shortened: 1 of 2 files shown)


metaodi:
[The tests](https://github.com/ckan/ckanext-harvest/blob/7b251f0e3b26e9567025b833228676b8801979d0/ckanext/harvest/tests/test_action.py#L696-L736) only use HarvestGatherError's, so I guess you should add a test with HarvestObjectError as well.

stefina:
To add tests for this case definitely makes sense, I hope to get this done sooner than later.

PR #385, opened 2019-12-17
DGU fixes

kentsanggds:
## What

Added a number of fixes and additional features to enable data.gov.uk to work better

These have been summarised as 

- added a `not modified` status, as the counts were confusing for harvest jobs
- stop sending the job to the gather queue if running harvest job manually with `run_test`
- prevent Redis failures from causing the stack to break by capturing the errors from Redis
- remove harvest_source statistics as the long running queries can break the stack
- add a command to allow `force-import` of datasets to make it easier to target datasets for testing

## Why

These commits have made the CKAN stack more stable for data.gov.uk, hopefully they will also improve the resilience for other CKAN stacks and make it easier for other devs to work on harvest issues.


amercader:
Thanks @kentsanggds, this all looks helpful. I've removed the `overall_statistics` completely as they were not used at all and we can always re-implement them better in the future if necessary.

PR #389, opened 2020-01-22
Check for auth before rendering admin link.

starsinmypockets:
Note that the incorectly rendered admin link does _not_ grant authorization to update the harvest source. The auth appears to be correct. The problems appears to be that the admin link is rendered when it should not be. Following the link results in 401 when rendered for a user without the correct permissions.


amercader:
Good catch @starsinmypockets

PR #390, opened 2020-02-01
Don't swallow exceptions

starsinmypockets:
Remove blanket exceptions from auth functions.


amercader:
This is a good move. If there are errors raised, they need to be fixed.

PR #397, opened 2020-03-19
fix clearsource_history command to avoid disassociation

mutantsan:
We encountered a huge problem with disassociation on a production serer a few days ago. This PR tries to fix it by adding an option to an existing clearsource_history command. The main idea is when we have the harvest_object associated with an existing package, we don't wanna loose it and so we don't wanna delete the harvest_job. I'm not entirely sure about this PR, so I'll appreciate any observation about this approach.

### DIFF
--- a/ckanext/harvest/cli.py
+++ b/ckanext/harvest/cli.py
@@ -108,7 +108,11 @@ def clear(ctx, id):
 
 @source.command()
 @click.argument(u"id", metavar=u"SOURCE_ID_OR_NAME", required=False)
-@click.argument(u"keep_actual", default=False, required=False)
+@click.option(
+    "-k",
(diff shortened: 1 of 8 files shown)


seitenbau-govdata:
Add references to #268 and #393

seitenbau-govdata:
We plan to open a new pull request based on this pull request to add an option to keep the latest related objects to a harvest source. This will probably be happen within the next three weeks.

seitenbau-govdata:
@mutantsan Thanks for your input on which we have created our PR https://github.com/ckan/ckanext-harvest/pull/484 and finished the work.

mutantsan:
@seitenbau-govdata thanks, I appreciate it!

PR #398, opened 2020-03-26
Abort failed jobs cli command

mutantsan:
Sometimes, we encounter the situation when job is in a "limbo state". Situation, when the job has run with errors but the harvester run command will not mark it as finished, and therefore you cannot run another job. This PR comes to abort such jobs if they are runs too long (depends on source update frequency). We planning to run it right before harvest run, to be able reharvest source if it's job ends up with error.

We used clearsource_history command to get rid of hanged jobs before and it was erroneous decision.

For example, `paster harvester abort_failed_jobs 3 -c /etc/ckan/default/production.ini` going to abort all jobs working longer then it's source update frequency (daily, weekly, monthly etc) or if update frequency is manual or always, longer then 3 days.

### DIFF
--- a/ckanext/harvest/cli.py
+++ b/ckanext/harvest/cli.py
@@ -206,18 +206,18 @@ def job_abort(ctx, id):
     "-e",
     "--exclude",
     default=False,
-    help="""If source_id provided as excluded, all sources failed jobs, except for that 
+    help="""If source_id provided as excluded, all sources failed jobs, except for that
     will be aborted. You can use comma as a separator to provide multiple source_id's""",
 )
 @click.pass_context
 def abort_failed_jobs(ctx, life_span, include, exclude):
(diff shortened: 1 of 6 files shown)


mutantsan:
Seems like test failure doesn't related to my changes

amercader:
Thanks @mutantsan this looks really useful, but could you check if it is relevant or it plays nice with the changes on #403 (the job timeout feature). I think they are both compatible but jsut want to double check.

mutantsan:
@amercader I've tested it again and it's seems to work for me as expected.
So, if I understand in right, in #403  we have a new config option `ckan.harvest.timeout` and it's great. But a source with 50 datasets and a source with 50k datasets should run for a completely different time. What if we have 10 custom harvesters and 20 sources.
In this PR we can include and exclude sources for check by `--include` and `--exclude` options.

PR #399, opened 2020-03-27
Travis fix for Postgres 10

davidread:
Without this PR, when testing CKAN 2.6 and 2.7 these days, you get [this
error on Travis](https://travis-ci.org/github/ckan/ckanext-harvest/jobs/667788749):
```
  Collecting psycopg2==2.4.5 (from -r requirements.txt (line 34))
  Downloading https://files.pythonhosted.org/packages/36/77/894a5dd9f3f55cfc85682d3e6473ee5103d8d418b95baf4019fad3ffa026/psycopg2-2.4.5.tar.gz (719kB)
    100% |████████████████████████████████| 727kB 2.2MB/s
    Complete output from command python setup.py egg_info:
    running egg_info
    creating pip-egg-info/psycopg2.egg-info
    writing pip-egg-info/psycopg2.egg-info/PKG-INFO
    writing top-level names to pip-egg-info/psycopg2.egg-info/top_level.txt
    writing dependency_links to pip-egg-info/psycopg2.egg-info/dependency_links.txt
    writing manifest file 'pip-egg-info/psycopg2.egg-info/SOURCES.txt'
    Error: could not determine PostgreSQL version from '10.1'
```
I'm not quite sure why Postgres 10 is found, because the image claims to be running Postgres 9.2 in the logs:
```
Pre-installed PostgreSQL versions
9.2.24
9.3.20
9.4.15
9.5.10
9.6.6
...
# line 422ish
 * Stopping PostgreSQL 9.2 database server        * Stopping PostgreSQL 9.3 database server        * Stopping PostgreSQL 9.4 database server        * Stopping PostgreSQL 9.5 database server        * Stopping PostgreSQL 9.6 database server        * Starting PostgreSQL 9.2 database server       sudo service postgresql start 9.2
```
And I don't know why this problem hasn't reared on ckanext-harvest builds before, although there hasn't been one for a month. I've had to do this hack on other extensions.

But whatever the reason, this PR fixes it 🤷‍♂

### DIFF
--- a/bin/travis-build.bash
+++ b/bin/travis-build.bash
@@ -31,6 +31,20 @@ fi
 
 python setup.py develop
 
+if (( $CKAN_MINOR_VERSION < 8 )) && (( $PYTHON_MAJOR_VERSION == 2 ))
+then
+    echo "Hack to get CKAN 2.6 / 2.7 to work with Postgres 10..."
+    # because Postgres 10 is what Travis' Ubuntu image gives us now
+    sed -i -e 's/psycopg2==.*/psycopg2==2.8.2/' requirements.txt
+    sed -i -e 's/except sqlalchemy.exc.InternalError:/except (sqlalchemy.exc.InternalError, sqlalchemy.exc.DBAPIError):/' ckan/config/environment.py
+    if (( $CKAN_MINOR_VERSION == 7 ))
+    then
+        sed -i -e 's/ connection.connection,/ connection.connection.connection,/' ckanext/datastore/backend/postgres.py
+    else
+        sed -i -e 's/ connection.connection,/ connection.connection.connection,/' ckanext/datastore/db.py
+    fi
+fi
+
 if (( $CKAN_MINOR_VERSION >= 9 )) && (( $PYTHON_MAJOR_VERSION == 2 ))
 then
     pip install -r requirements-py2.txt


Zharktas:
@davidread travis has a new feature for config validation (which can be turned off in project settings). The validation changes the group from deprecated-2017Q4 to deprecated-2017q4 which is a different image with build times  Tue Aug 29 02:48:34 UTC 2017 compared to Mon Jan  8 21:40:54 UTC 2018. The newer image has newer libpg which is the reason for the failure.

We really should settle to single solution, either use the old image or update the images and update psycopg during build. This affects pretty much every extension which has enabled travis builds recently.

davidread:
Thanks for the explanation @Zharktas  - most helpful!

It's the second time Travis has done something to stop us using these old versions, so it seems like skating on thin ice. We'll stop testing against CKAN 2.6 shortly, with the release of CKAN 2.9, but we'll probably want to test against CKAN 2.7 for the rest of this year. I'm mildly favouring this the hack in this PR and doing a batch of PRs against other extensions. At least it's in code so you can see what's going on, rather that in some Travis UI option that the extension owner has to 'know' to set.

I suppose the other option is to do a patch release for 2.7 updating the psycopg version. We could change the psycog version in requirements.txt but upgrading it is optional - i.e. the code works with either version. We don't normally add enhancements to patch releases, but if they are going out anyway, maybe it's easy to just include this too? Or maybe it's overkill to even involve users in this issue, when it's just our CI that needs fixing.

amercader:
I've spent some time cleaning up the travis setup and this is no longer relevant

PR #403, opened 2020-05-01
Add support for Redis 2.10.5 in CKAN 2.7

kentsanggds:
## What 

CKAN 2.7 uses Redis 2.10.5

https://github.com/ckan/ckan/blob/9a6e9cdbd5a9fc8004d472a3d0076ecb3feda01a/requirements.txt#L42

but the latest CKAN uses Redis 3.3.11 which has changed how args are called.

https://github.com/ckan/ckan/blob/master/requirements.txt#L42

This will support both types of call.

Also added in ability to filter the harvest source list by organization id, and to limit it by default to 100, this is configurable in the config file, as this was causing us timeouts.

And an ability to timeout harvest jobs as we were experiencing some very long running harvest jobs with no idea why they were hanging as the error messages were not available. This is only active if the configuration option `ckan.harvest.timeout` is set.


ckanext/harvest/queue.py
```diff
@@ -203,7 +203,13 @@ def send(self, body, **kw):
         value = json.dumps(body)
         # remove if already there
         if self.routing_key == get_gather_routing_key():
-            self.redis.lrem(self.routing_key, 0, value)
+            # it appears that both types of call are possible within the redis library depending on which version used
+            # for now support both versions
+            # https://github.com/andymccurdy/redis-py#client-classes-redis-and-strictredis
+            try:
+                self.redis.lrem(self.routing_key, 0, value)
+            except:
```
amercader, line 211: Can you capture the specific exception that the wrong version causes, rather than a blanket except? Cheers!

amercader:
Thanks @kentsanggds , in the future please split separate features or fixes in separate PRs as it makes the review easier.
The `harvest_source_list` action should be completely rewritten to use `package_search` instead of the old DB models, but your changes are a good addition in the meantime.

kentsanggds:
Yes, sorry about all the extra commits, didn't realise that a PR on master against upstream would merge everything, will split them out into another branch for merging upstream in future

PR #407, opened 2020-06-09
Improve harvest jobs notifications

### LINKED ISSUE #406: Improve notifications
As an admin of an organization that relies heavily on harvesting, I want the option to receive job completion summaries for all jobs (not just errors), so that I'm notified of **all** changes related to harvest jobs.

I think it would be helpful for some users of this extension to have the option to enable summary email notifications, which would be very similar to the error email, only with some extra information (records added, updated, deleted, and a list of those packages). 

A simple implementation of this _could_ be: rename `send_error_mail` to something like `send_summary_mail` or `send_notification` (and make it a universal report template, or we can create a new function for summaries), add the extra records information, and add a new `.env` variable option like `ckanext.harvest.status_mail.summary` (leaving the option for error only mail as well).

**Tasks**
 - [x] Add a new config value to send notifications for all jobs
 - [x] Rename the _send_error_mail_ and use it for all jobs (if we set this up)
 - [x] Add test to cover this change  

**Notes**
I already start working on this.

avdata99:
Fixes #406 (and related to [GSA#315](https://github.com/GSA/datagov-ckan-multi/issues/315))

 - Added new config `ckan.harvest.status_mail.all` to notify all finished jobs.
   - README updated
 - Text templates were added for emails and variables were configured to fill it
 - Tests updated and [green at CI](https://travis-ci.org/github/ckan/ckanext-harvest/builds/696512648) for CKAN 2.6, 2.7 and 2.8. Master is not working (as in `master`)
 - Allow plugins to add custom recipients by adding the `IAction: add_extra_notification_recipients` [here](https://github.com/avdata99/ckanext-harvest/blob/406_notifications/ckanext/harvest/logic/action/update.py#L737-L745)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -191,7 +191,7 @@ If you want to send an email when a **Harvest Job fails**, you can set the follo
 
     ckan.harvest.status_mail.errored = True
 
-If you want to send an email when **all Harvest Jobs finish** (whether or not it failed), you can set the following configuration option in the ini file:
+If you want to send an email when **completed Harvest Jobs finish** (whether or not it failed), you can set the following configuration option in the ini file:
 
     ckan.harvest.status_mail.all = True
 
--- a/README.rst
+++ b/README.rst
@@ -191,7 +191,7 @@ If you want to send an email when a **Harvest Job fails**, you can set the follo
 
     ckan.harvest.status_mail.errored = True
 
-If you want to send an email when **all Harvest Job finishes** (whether or not it failed), you can set the following configuration option in the ini file:
+If you want to send an email when **all Harvest Jobs finish** (whether or not it failed), you can set the following configuration option in the ini file:
 
     ckan.harvest.status_mail.all = True
 
--- a/README.rst
+++ b/README.rst
@@ -187,9 +187,13 @@ If you don't specify this setting, the default will be number-sequence.
 Send error mails when harvesting fails (optional)
 =================================================
 
-If you want to send an email when a Harvest Job fails, you can set the following configuration option in the ini file:
+If you want to send an email when a **Harvest Job fails**, you can set the following configuration option in the ini file:
 
     ckan.harvest.status_mail.errored = True
+
+If you want to send an email when **all Harvest Job finishes** (whether or not it failed), you can set the following configuration option in the ini file:
+
+    ckan.harvest.status_mail.all = True
 
 That way, all CKAN Users who are declared as Sysadmins will receive the Error emails at their configured email address. If the Harvest-Source of the failing Harvest-Job belongs to an organization, the error-mail will also be sent to the organization-members who have the admin-role if their E-Mail is configured.
 
--- a/README.rst
+++ b/README.rst
@@ -881,6 +881,35 @@ following steps with the one you are using.
    This particular example will perform clean-up each day at 05 AM.
    You can tweak the value according to your needs.
 
+Extensible actions
+==================
+
+Recipients on harvest jobs notifications
+----------------------------------------
+
+:code:`harvest_get_notifications_recipients`: you can *chain* this action from another extension to change 
+the recipients for harvest jobs notifications.
+
+.. code-block:: python
+
+  @toolkit.chained_action
+  def harvest_get_notifications_recipients(up_func, context, data_dict):
+      """ Harvester plugin notify by default about harvest jobs only to 
+              admin users of the related organization.
+              Also allow to add custom recipients with this function.
+              
+          Return a list of dicts with name and email like
+              {'name': 'John', 'email': 'john@source.com'} """
+
+      recipients = up_func(context, data_dict)
+      new_recipients = []
+
+      # you custom logic to add new_recipients here
+      # new_recipients.append({'name': 'Harvester Admin', 'email': 'admin@harvester-team.com'})
+      # recipients += new_recipients
+      return recipients
+
+
 Tests
 =====
 
@@ -900,6 +929,7 @@ Here are some common errors and solutions:
 
 * ``(OperationalError) near "SET": syntax error``
   You are testing with SQLite as the database, but the CKAN Harvester needs PostgreSQL. Specify test-core.ini instead of test.ini.
+
 
 Harvest API
 =====
--- a/ckanext/harvest/logic/action/update.py
+++ b/ckanext/harvest/logic/action/update.py
@@ -39,6 +39,7 @@ from ckanext.harvest.logic.action.get import (
 
 import ckan.lib.mailer as mailer
 from itertools import islice
+from ckan.lib.base import render_jinja2
 
 log = logging.getLogger(__name__)
 
@@ -660,11 +661,22 @@ def harvest_jobs_run(context, data_dict):
                     get_action('harvest_source_reindex')(
                         context, {'id': job_obj.source.id})
 
-                    status = get_action('harvest_source_show_status')(context, {'id': job_obj.source.id})
+                    status = get_action('harvest_source_show_status')(
+                        context, {'id': job_obj.source.id})
 
-                    if toolkit.asbool(config.get('ckan.harvest.status_mail.errored'))\
-                            and (status['last_job']['stats']['errored']):
-                        send_error_mail(context, job_obj.source.id, status)
+                    notify_all = toolkit.asbool(config.get('ckan.harvest.status_mail.all'))
+                    notify_errors = toolkit.asbool(config.get('ckan.harvest.status_mail.errored'))
+                    last_job_errors = status['last_job']['stats'].get('errored', 0)
+
+                    if notify_all:
+                        subject, body = prepare_summary_mail(context, job_obj.source.id, status)
+                        log.info('Sending summary email')
+                        send_mail(context, job_obj.source.id, subject, body)
+                    elif notify_errors and last_job_errors > 0:
+                        subject, body = prepare_error_mail(context, job_obj.source_id, status)
+                        log.info('Sending error mail')
+                        send_mail(context, job_obj.source.id, subject, body)
+                    
                 else:
                     log.debug('Ongoing job:%s source:%s',
                               job['id'], job['source_id'])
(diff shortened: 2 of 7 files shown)


README.rst
```diff
@@ -187,10 +187,14 @@ If you don't specify this setting, the default will be number-sequence.
 Send error mails when harvesting fails (optional)
 =================================================
 
-If you want to send an email when a Harvest Job fails, you can set the following configuration option in the ini file:
+If you want to send an email when a **Harvest Job fails**, you can set the following configuration option in the ini file:
 
     ckan.harvest.status_mail.errored = True
 
+If you want to send an email when **all Harvest Job finishes** (whether or not it failed), you can set the following configuration option in the ini file:
```
thejuliekramer, line 194:
```suggestion
If you want to send an email when **a Harvest Job finishes** (whether or not it failed), you can set the following configuration option in the ini file:
```

thejuliekramer, line 194:
```suggestion
If you want to send an email when **all Harvest Jobs finish** (whether or not it failed), you can set the following configuration option in the ini file:
```

ckanext/harvest/logic/action/update.py
```diff
@@ -605,106 +622,136 @@ def harvest_jobs_run(context, data_dict):
     return []  # merely for backwards compatibility
 
 
-def send_error_mail(context, source_id, status):
```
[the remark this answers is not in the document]
avdata99: I just sent a [new commit](https://github.com/ckan/ckanext-harvest/pull/407/commits/fd3ffd1b65382b701ce2306c3e8de478c2b2ce50) to recover this function as Adria recommends in the issue's comments.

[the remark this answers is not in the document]
thejuliekramer: It looks like we are still removing the definition of send_error_mail here

README.rst
```diff
@@ -187,10 +187,14 @@ If you don't specify this setting, the default will be number-sequence.
...
 =================================================
 
-If you want to send an email when a Harvest Job fails, you can set the following configuration option in the ini file:
+If you want to send an email when a **Harvest Job fails**, you can set the following configuration option in the ini file:
 
     ckan.harvest.status_mail.errored = True
 
+If you want to send an email when **all Harvest Jobs finish** (whether or not it failed), you can set the following configuration option in the ini file:
+
+    ckan.harvest.status_mail.all = True
```
adborden, line 196: "completed" might be a better term than all, since we're notifying on any job completion regardless of status.

ckanext/harvest/logic/action/update.py
```diff
@@ -586,11 +588,21 @@ def harvest_jobs_run(context, data_dict):
...
 
-                    if toolkit.asbool(config.get('ckan.harvest.status_mail.errored'))\
-                            and (status['last_job']['stats']['errored']):
-                        send_error_mail(context, job_obj.source.id, status)
+                    notify_all = toolkit.asbool(config.get('ckan.harvest.status_mail.all'))
+                    notify_errors = toolkit.asbool(config.get('ckan.harvest.status_mail.errored'))
+                    last_job_errors = status['last_job']['stats'].get('errored', 0)
+                    log.info('Notifications: All:{} On error:{} Errors:{}'.format(notify_all, notify_errors, last_job_errors))
+                    
+                    if notify_all:
```
adborden, line 599: The duplicated `send_error_email` is a bit hard to follow. How about inverting the conditions to be about the error status vs success status? Something like:

```python
if last_job_errors > 0:  # completed with error
  if notify_all or notify_error:
    send_error_email(context, job_obj.source_id, status)
else: # completed succesfully
  if notify_all:
    send_summary_email(context, job_obj.source.id, status)
```

or (maybe too compact):

```python
if last_job_errors > 0 and (notify_all or notify_error):
  send_error_email(context, job_obj.source_id, status)
elif notify_all:
  send_summary_email(context, job_obj.source.id, status)
```

ckanext/harvest/logic/action/update.py
```diff
@@ -39,6 +38,7 @@
 
 import ckan.lib.mailer as mailer
 from itertools import islice
+from ckan.lib.base import render_jinja2
```
amercader, line 41: Is it possible to use `toolkit.render` here?

ckanext/harvest/logic/action/update.py
```diff
@@ -586,11 +586,18 @@ def harvest_jobs_run(context, data_dict):
...
+                    status = get_action('harvest_source_show_status')(
+                        context, {'id': job_obj.source.id})
 
-                    if toolkit.asbool(config.get('ckan.harvest.status_mail.errored'))\
-                            and (status['last_job']['stats']['errored']):
-                        send_error_mail(context, job_obj.source.id, status)
+                    notify_all = toolkit.asbool(config.get('ckan.harvest.status_mail.all'))
+                    notify_errors = toolkit.asbool(config.get('ckan.harvest.status_mail.errored'))
+                    last_job_errors = status['last_job']['stats'].get('errored', 0)
+                    log.info('Notifications: All:{} On error:{} Errors:{}'.format(notify_all, notify_errors, last_job_errors))
```
amercader, line 595: Can you turn this into `log.debug`? The extension is already quite verbose :)

ckanext/harvest/logic/action/update.py
```diff
@@ -605,106 +612,156 @@ def harvest_jobs_run(context, data_dict):
...
+
+    # allow plugins to add custom recipients 
+    for p in PluginImplementations(IActions):
+        actions = p.get_actions()
+        if 'add_extra_notification_recipients' in actions.keys():
+            fn = actions['add_extra_notification_recipients']    
+            data_dict = {'source_id': source_id}
+            new_recipients = fn(context, data_dict)
+            if len(new_recipients) > 0:
+                log.info('New recipients added to notification: {}'.format(new_recipients))
```
amercader, line 744: log.debug

ckanext/harvest/logic/action/update.py
```diff
@@ -605,106 +612,156 @@ def harvest_jobs_run(context, data_dict):
...
+                'user_show')(context, {'id': member[0]})
+
+            if member_details['email']:
+                recipients.append({
+                    'name': member_details['name'],
+                    'email': member_details['email']
+                })
+
+    # allow plugins to add custom recipients 
+    for p in PluginImplementations(IActions):
```
amercader, line 737: The standard way of extending actions is to use [chained actions](https://docs.ckan.org/en/2.9/extensions/plugin-interfaces.html#ckan.plugins.interfaces.IActions.get_actions). ckanext-harvest would implement the main `harvest_add_extra_notification_recipients` (better to prefix it with `harvest_` to avoid conflicts), which would replace the current `get_recipients`.
Then other extension would implement their own action as:

```python
@toolkit.chained_action
def harvest_add_extra_notification_recipients(up_func, context, data_dict):
    recipients = up_func(context, data_dict)
    # Custom logic
    recipients += new_recipients
    return recipients
```

Does this make sense?

amercader (commented):
@avdata99 sorry it took me a while to review this. This is looking great, I just left some minor comments.
Could you also please merge the latest master? All tests should be green now.

ckanext/harvest/logic/action/update.py
avdata99: Done [here](https://github.com/ckan/ckanext-harvest/pull/407/commits/644316ed1781f4a1250342b360135cbad6a9081e)

ckanext/harvest/logic/action/update.py
avdata99: Done [here](https://github.com/ckan/ckanext-harvest/pull/407/commits/644316ed1781f4a1250342b360135cbad6a9081e)

ckanext/harvest/logic/action/update.py
avdata99: @amercader we are [using this action](https://github.com/GSA/ckanext-datagovcatalog/blob/master/ckanext/datagovcatalog/harvester/notifications.py#L9) from a custom extension (not a harvester). This extension analyzes `extras` from an `organization` to look for new recipients.
So, maybe this is not only for harvesters. Probably users will want to define extra recipients in configuration or in other extensions. I think _recipients_ is not just about harvesters

Anyway, [I try to create a new branch](https://github.com/avdata99/ckanext-harvest/commit/f59e373341cb31c2ba340cc1d340d64963a4670c) to transform this into a `chained_action`. Is this a good approach? I'm not sure.
If we use this: will we need to use in all harvester or we will be able to use from a non-harvester extension?

ckanext/harvest/logic/action/update.py
avdata99: Done [here](https://github.com/avdata99/ckanext-harvest/commit/ba9186015abb34ad225c34c70ee0310c240d134f)

ckanext/harvest/logic/action/update.py
amercader: @avdata99 I think actions should be explicit in what they do and be granular. So my suggested approach would be:

In **ckanext-harvest**:

In *action/get.py*:

```python
# This will automatically get registered as action, no need to explicitly register it in plugin.py
def harvest_get_notifications_recipients(context, data_dict):
   # See auth/get.py below
   check_access('harvest_get_notifications_recipients', contex, data_dict)

   # Include here the default logic for getting recipients (what you have in this PR minus the lines below to call other actions)
    return recipients

```

In *auth/get.py*:

```python

def harvest_get_notifications_recipients(context, data_dict):
   # Only sysadmins can access this
    return {'success': False}
```

In *action/update.py*:

```python

def send_summary_email(context, source_id, status):
    subject, body = prepare_summary_mail(context, source_id, status)
    # Using get_action here will ensure that if another extension implements this action it will be called (if loaded first in the ini file)
    recipients = toolkit.get_action('harvest_get_notifications_recipients')(context, {'source_id': source_id})
    send_mail(recipients, subject, body)

def send_error_email(context, source_id, status):
    subject, body = prepare_error_mail(context, source_id, status)
    recipients = toolkit.get_action('harvest_get_notifications_recipients')(context, {'source_id': source_id})
    send_mail(recipients, subject, body)

```

In **ckanext-datagovcatalog**:

In *actions.py* (or whatever file depending on how you structured your extension):

Option 1, if you want to extend the default behaviour from ckanext-harvest, use a chained action:

```python
@toolkit.chained_action
def harvest_get_notifications_recipients(up_func, context, data_dict):
     # This calls the ckanext-harvest parent function
     recipients = up_func(context, data_dict)

    # Do any extra stuff to modify the default recipients
    
     return recipients
```

Option 2, If you want another behaviour, just override the whole function:
```python
def harvest_get_notifications_recipients(context, data_dict):

    # Do your custom stuff
    
     return recipients
```

If you have a generic `get_recipients()` function which is used elsewhere on your extension for other notifcations just call it from the overridden action:

 ```python
def harvest_get_notifications_recipients(context, data_dict):

    recipients = get_recipients()
    
     return recipients
```

Does this makes sense? I think that makes the implementation simpler and cleaner

avdata99: Ok @amercader, I make these changes [here](https://github.com/ckan/ckanext-harvest/pull/407/commits/c84f63896c6e483e17df9fc10c6d22ff6ec94dd5).
I will update my extension to fit this

amercader:
@avdata99 Great! One last thing before merging, can you document the new `harvest_get_notifications_recipients` action in the README and how it can be extended from other plugins? Thanks

avdata99:
> @avdata99 Great! One last thing before merging, can you document the new `harvest_get_notifications_recipients` action in the README and how it can be extended from other plugins? Thanks

I just sent [this comments](https://github.com/ckan/ckanext-harvest/pull/407/commits/b798ded3d4420a6ef813bb9c9ac285291997cbb9) to the README file.

amercader:
Thanks for this @avdata99 !

PR #421, opened 2020-10-07
Fix checking for harvest objects in fetch queue.

### LINKED ISSUE #420: Bug causing Redis fetch queue to grow constantly
Recently, we noticed that our harvesters were showing very poor performance, with jobs that should take less than an hour to complete taking several days. I tracked the problem down to the following commit in this ckanext: https://github.com/ckan/ckanext-harvest/commit/5ffe6d4bfa8d6e33a25a5c73dc57c241d59ead17

Ckanext-harvest stores harvest object ids in the fetch queue in a list in redis, under the key `'ckanext-harvest:{site_id}:harvest_object_id'`. The commit adds a method `resubmit_objects` to get the ids of all harvest objects in the `WAITING` state from the database, and check whether they are in the Redis fetch queue. The problem is that for each object it checks whether the object id is a key in Redis, and when it isn't, it re-sends the harvest object to Redis. The harvest object id is added to the list, as usual, even though it might already be in it.

`resubmit_objects` is called every time the `harvester run` command is run, which can be frequently.

As an example, our CKAN instance runs `harvester run` every five minutes. We have 7700 datasets on the instance. I purged the queues about 24 hours ago and right now there are 1417987 harvest object ids in the fetch queue!

bellisk:
The fetch queue is stored in redis as a list of harvest object ids. The previous code checked for individual harvest object ids in redis, and always found none. Therefore, every time the harvester run command was run, all ids of harvest objects in the WAITING state in the DB were re-added to the fetch queue.

Fixes #420.

### DIFF
--- a/ckanext/harvest/queue.py
+++ b/ckanext/harvest/queue.py
@@ -167,8 +167,18 @@ def resubmit_objects():
         .filter_by(state='WAITING') \
         .all()
 
+    objects_in_queue = []
+    fetch_routing_key = get_fetch_routing_key()
+
+    for i in range(redis.llen(fetch_routing_key)):
+        item = redis.lpop(fetch_routing_key)
+        objects_in_queue.append(
+            json.loads(item)['harvest_object_id']
+        )
+        redis.rpush(fetch_routing_key, item)
+
     for object_id, in waiting_objects:
-        if not redis.get(object_id):
+        if object_id not in objects_in_queue:
             log.debug('Re-sent object {} to the fetch queue'.format(object_id))
             publisher.send({'harvest_object_id': object_id})
 
--- a/ckanext/harvest/queue.py
+++ b/ckanext/harvest/queue.py
@@ -170,12 +170,8 @@ def resubmit_objects():
     objects_in_queue = []
     fetch_routing_key = get_fetch_routing_key()
 
-    for i in range(redis.llen(fetch_routing_key)):
-        item = redis.lpop(fetch_routing_key)
-        objects_in_queue.append(
-            json.loads(item)['harvest_object_id']
-        )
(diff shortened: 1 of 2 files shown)


ckanext/harvest/queue.py
```diff
@@ -167,8 +167,18 @@ def resubmit_objects():
...
 
+    objects_in_queue = []
+    fetch_routing_key = get_fetch_routing_key()
+
+    for i in range(redis.llen(fetch_routing_key)):
+        item = redis.lpop(fetch_routing_key)
+        objects_in_queue.append(
+            json.loads(item)['harvest_object_id']
+        )
+        redis.rpush(fetch_routing_key, item)
```
amercader, line 178: Instead of popping and re-pushing each item from Redis we can get them all in one go and process them in Python:

```suggestion
    for item in redis.lrange(fetch_routing_key, 0, -1):
        objects_in_queue.append(
            json.loads(item)['harvest_object_id']
        )
```

Or in one line if you are feeling fancy:

```python
objects_in_queue = [json.loads(o)['harvest_object_id'] for o in redis.lrange(fetch_routing_key, 0, -1)]
```

amercader:
Thanks a lot @bellisk , this was a bit embarrassing on my side, thanks for spotting it.
I just added a minor suggestion, if you can make the change I'll merge and release a new version as this is quite bad

ckanext/harvest/queue.py
bellisk: Thanks for the improvement! I'll make that change right now. :)

bellisk:
Ready to merge, @amercader. :) Thanks for looking at it so quickly.

amercader:
I've just release 1.3.2 including this fix, thanks again @bellisk !

Issue #425, opened 2020-11-11
Extend harvester to support include/exclude filtering by tags


scottlimmer:
I have a need to harvest records from another CKAN instance filtered by Organisation and Tag. 

Planning to replicate organization/groups include/exclude code to allow tag filtering. Logging issue to go with forthcoming pull request as I figure this might be useful to others.

PR #427, opened 2020-11-12
[#425] Extend harvester to support include/exclude filtering by tags

scottlimmer:
Added ability to include/exclude filter by tags as per organization/groups.
PR includes logic, CHANGELOG, README, tests


rickelt:
Is anybody still working on this issue? 
I'm interested in using that feature with CKAN 2.9

PR #437, opened 2021-03-12
Optimize last error free job

Zharktas:
last_error_free_job executes select query every time it accesses harvest objects belonging to the harvest job as they are lazily loaded. This results in inefficient cpu and memory use when there are lot of harvest jobs with errors in them.

This PR does eager query for harvest objects and does the filtering in the same query. If no harvest objects are returned for the job, the job is error free.

### DIFF
--- a/ckanext/harvest/harvesters/base.py
+++ b/ckanext/harvest/harvesters/base.py
@@ -5,7 +5,7 @@ import re
 import uuid
 import six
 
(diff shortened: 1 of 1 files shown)


bzar (approved):

tino097:
@Zharktas can you please fix the conflict?

Zharktas:
@tino097 done

PR #438, opened 2021-03-31
Move adding harvest extras to before_index

chris-macdermaid:


amercader:
@chris-macdermaid Thanks for this, it looks good. It just needs a bit of tweaking on how the dict is modified in the `before_index()` hook. The dict that this hook gets is not a standard dataset dict that gets returned by `package_show` but rather the payload that gets sent to Solr for indexing. We need to change the following keys:

```python
{
   'name': 'my-dataset',
   'data_dict': '{...}',  # This is JSON dump of the dataset dict before validation
   'validated_data_dict': '{...}', # This is JSON dump of the dataset dict after validation
   'extras_harvest_source_id': 'xxx'  # There is one of these fields for each extra, which will get index by Solr so you can filter, facet, etc by them
}
```

So in `before_index()` we should set the harvest extras in these three places. Here's a quick patch I put together based on your branch. Can you test it and see if the datasets are indexed as expected (ie the harvest fields are returned)?

```diff
diff --git a/ckanext/harvest/plugin/__init__.py b/ckanext/harvest/plugin/__init__.py
index 4349c0a..b7e2742 100644
--- a/ckanext/harvest/plugin/__init__.py
+++ b/ckanext/harvest/plugin/__init__.py
@@ -111,34 +111,32 @@ class Harvest(MixinPlugin, p.SingletonPlugin, DefaultDatasetForm, DefaultTransla
 
     def before_index(self, pkg_dict):
 
-        if 'type' not in pkg_dict or pkg_dict['type'] != DATASET_TYPE_NAME:
-            # This is a normal dataset, check if it was harvested and if so, add
-            # info about the HarvestObject and HarvestSource
-
-            harvest_object = model.Session.query(HarvestObject) \
-                    .filter(HarvestObject.package_id == pkg_dict['id']) \
-                    .filter(HarvestObject.current == True).first() # noqa
-
-            # If the harvest extras are there, remove them. This can happen eg
-            # when calling package_update or resource_update, which call
-            # package_show
-            if pkg_dict.get('extras'):
-                pkg_dict['extras'][:] = [e for e in pkg_dict.get('extras', [])
-                                         if not e['key']
-                                         in ('harvest_object_id', 'harvest_source_id', 'harvest_source_title',)]
-            else:
-                pkg_dict['extras'] = []
-            # We only want to add these extras at index time so they are part
-            # of the cached data_dict used to display, search results etc. We
-            # don't want them added when editing the dataset, otherwise we get
-            # duplicated key errors.
-            if harvest_object:
-                for key, value in [
-                    ('harvest_object_id', harvest_object.id),
-                    ('harvest_source_id', harvest_object.source.id),
-                    ('harvest_source_title', harvest_object.source.title),
-                        ]:
-                    _add_extra(pkg_dict, key, value)
+        harvest_object = model.Session.query(HarvestObject) \
+                .filter(HarvestObject.package_id == pkg_dict['id']) \
+                .filter(HarvestObject.current == True).first() # noqa
+
+        if harvest_object:
+
+            data_dict = json.loads(pkg_dict['data_dict'])
+
+            validated_data_dict = json.loads(pkg_dict['validated_data_dict'])
+
+            harvest_extras = [
+                ('harvest_object_id', harvest_object.id),
+                ('harvest_source_id', harvest_object.source.id),
+                ('harvest_source_title', harvest_object.source.title),
+            ]
+
+            for key, value in harvest_extras:
+
+                data_dict['extras'].append({'key': key, 'value': value})
+
+                validated_data_dict['extras'].append({'key': key, 'value': value})
+
+                pkg_dict[f'extras_{key}'] = value
+
+            pkg_dict['data_dict'] = json.dumps(data_dict)
+            pkg_dict['validated_data_dict'] = json.dumps(validated_data_dict)
 
         return pkg_dict
 
@@ -350,18 +348,6 @@ class Harvest(MixinPlugin, p.SingletonPlugin, DefaultDatasetForm, DefaultTransla
                             ])
 
 
-def _add_extra(data_dict, key, value):
-    if 'extras' not in data_dict:
-        data_dict['extras'] = []
-
-    if 'extras_' + key not in data_dict:
-        data_dict['extras_' + key] = []
-
-    data_dict['extras_' + key].append({
-        'key': key, 'value': value, 'state': u'active'
-    })
-
-
 def _get_logic_functions(module_root, logic_functions={}):
 
     for module_name in ['get', 'create', 'update', 'patch', 'delete']:

```

One thing worth checking if this still works as expected with custom schemas that move the harvest fields to the root level. I can test that once we validated the standard use case.

Let me know if all this makes sense

chris-macdermaid:
Related to #438

chris-macdermaid:
@amercader Thank you for the patch. Yes, this makes sense. I've added that to my PR. Working on testing locally.

Issue #428, opened 2020-12-23
gather_consumer -> gather-consumer; fetch_consumer -> fetch-consumer


kdwarn:
I'm using the latest version (1.3.2) of this plugin on CKAN 2.9, installed about a week ago. Following along with the instructions (both for manual runs and for setting this up in production), I had to change the command from `fetch_consumer` to `fetch-consumer` and `gather_consumer` to `gather-consumer` in order for them to run (otherwise was getting "no such command" error). I was going to submit a pull request with a change to the documentation to reflect this, but upon looking at the code, it looks like the original commands with the underscores are what the function names are. So I'm not sure where the change from `_` to `-` is taking place, or if perhaps there's something I'm missing about this and the reason I had to do it. I only realized that this was the solution when I ran `ckan harvester --help`, and the commands there (except for `harvesters_info`) are listed with `-` in them rather than `_`. Thoughts?

VladimirZD:
Hi, we just had this issue today, i lost about 1 hour before  we found out that gether-consumer is working and gether_consumer isnt...

Any ideas what is the cause of this?

seitenbau-govdata:
I think this isn't a issue in the code, because the terms `gather-consumer` and `fetch-consumer` cannot found in the repo by the GitHub search, whereas `gather_consumer` and `fetch_consumer` can found.

Perhaps you want to let us know your environments and show your scripts, so we can maybe see what is different?

amercader:
This is annoying. It looks like Click automatically changes underscores to hyphens https://github.com/pallets/click/issues/1123, so even though we define the command with an underscore to keep compatibility that gets ignored:

https://github.com/ckan/ckanext-harvest/blob/6222e38c85f04027f3fd9b2894a9c26ee7bd172f/ckanext/harvest/cli.py#L232

The fix is easy, just explicitly set the name with `@harvester.command("gather_consumer")`, but that would mean that we would break scripts that have used the new version and changed their scripts to `gather-consumer` in the meantime. And to be honest it would go against the convention in CKAN core as well so perhaps is better to bite to bullet and accept the change.

So my slight preference would be to leave things as they are, update the [README](https://github.com/ckan/ckanext-harvest#command-line-interface) and clearly state there that when upgrading to CKAN 2.9 you might need to adjust your scripts.

What do people think?

VladimirZD:
Updating README would help a lot, because that is first place ppl will look :).
[...]

PR #439, opened 2021-04-18
Update README -fix run-test command name

alantygel:
`run_test` command is actually spelled as `run-test`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -242,7 +242,7 @@ The following operations can be run from the command line as described underneat
         - clears all datasets, jobs and objects related to a harvest source,
           but keeps the source itself
 
-      harvester clearsource_history [{source-id}]
+      harvester clearsource-history [{source-id}]
         - If no source id is given the history for all harvest sources (maximum is 1000)
           will be cleared.
           Clears all jobs and objects related to a harvest source, but keeps the source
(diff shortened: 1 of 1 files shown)


amercader:
Thanks @alantygel . This is part of a wider change in the commands names in CKAN 2.9 (see https://github.com/ckan/ckanext-harvest/issues/428). Do you have time to fix the names for the other commands as well? All commands should replace the underscore with the hyphen:

```
Commands:
  abort-failed-jobs  Abort all jobs which are in a "limbo state" where the...
  clean-harvest-log  Clean-up mechanism for the harvest log table.
  fetch-consumer     Starts the consumer for the fetching queue.
  gather-consumer    Starts the consumer for the gathering queue.
  harvesters_info
  import             Perform the import stage with the last fetched
                     objects,...

  initdb             Creates the necessary tables in the database.
  job                Create new harvest job and runs it (puts it on the...
  job-abort          Marks a job as "Aborted" so that the source can be...
  job-all            Create new harvest jobs for all active sources.
  jobs               Lists harvest jobs.
  purge-queues       Removes all jobs from fetch and gather queue.
  reindex            Reindexes the harvest source datasets.
  run                Starts any harvest jobs that have been created by...
  run-test           Runs a harvest - for testing only.
  source             Manage harvest sources
  sources            Lists harvest sources.
```

If not no worries, I'll merge this as is. and we'll fix the rest later.

alantygel:
Sure!

PR #447, opened 2021-06-22
Added PY3 and CKAN2.9 support

korotkin:


Zharktas:
This makes no sense, harvest already is CKAN 2.9 and Py3 compatible and we still need to support older versions. Why these changes are made?

PR #452, opened 2021-08-02
Allow GET requests to harvester.refresh

bzar:
The "Reharvest" button in harvest source admin view returns a 404 because it performs a GET request and the blueprint doesn't cover them. Added GET to allowd methods to fix this.

Financed by Finland's open data portal opendata.fi. Find all Finnish open data at https://www.opendata.fi/en.
The Service is provided by the Digital and Population Data Services Agency (https://dvv.fi/en/).

### DIFF
--- a/ckanext/harvest/views.py
+++ b/ckanext/harvest/views.py
@@ -72,7 +72,7 @@ harvester.add_url_rule(
 )
 harvester.add_url_rule("/" + utils.DATASET_TYPE_NAME + "/clear/<id>",
                        view_func=clear,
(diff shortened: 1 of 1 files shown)


bzar:
The same issue seems to happen with clear

amercader:
I was wondering why this hasn't come up before as it seems pretty obvious and it's because the Reharvest and Clear buttons fire a modal dialog. When confirming it we must be sending a POST request with js instead of GET. It makes sense to support both.

@bzar can you apply the same fix to the Clear endpoint?

PR #458, opened 2021-09-20
Fix harvest extras for packages

jbrown-xentity:
It seems that injecting the harvest extra information at search/show time is very complex.
We (US data.gov CKAN team) propose a solution that puts the extra objects into the package itself, and remove the complex necessary code.

This was raised to our attention when we discovered that we had bad data from harvesting another CKAN source, and when the harvest information was no longer displayed when using `package_show`, only on `package_search` (which hurt our API users).

This replaces https://github.com/ckan/ckanext-harvest/pull/438, as our team member Chris has moved onto another project.

@amercader, any possibility we could merge this?

### DIFF
--- a/ckanext/harvest/plugin/__init__.py
+++ b/ckanext/harvest/plugin/__init__.py
@@ -79,26 +79,6 @@ class Harvest(MixinPlugin, p.SingletonPlugin, DefaultDatasetForm, DefaultTransla
             # Delete the actual HarvestSource object
             _delete_harvest_source_object(context, package_dict)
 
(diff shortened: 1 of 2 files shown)


jbrown-xentity:
Passing tests (in case they don't get run here): https://github.com/GSA/ckanext-harvest/actions/runs/1254660702

jbrown-xentity:
@amercader ping

amercader:
Many thanks @jbrown-xentity , this is a much simpler implementation

PR #472, opened 2021-10-22
Skip protected fields

### LINKED ISSUE #464: Protect specific metadata
It would be nice to be able to retain some previous metadata fields instead of dropping them; this would allow adding additional metadata from the CKAN interface on harvested resources, without risking losing them.

ckanext-harvest uses package update, but a switch to use package_patch instead should be pretty easy to implement. Otherwise, a list of protected metadata could be provided, so they can be copied over if needed.

frafra:
Fix #464.

### DIFF
--- a/ckanext/harvest/harvesters/base.py
+++ b/ckanext/harvest/harvesters/base.py
@@ -321,6 +321,9 @@ class HarvesterBase(SingletonPlugin):
                     package_dict.setdefault('name',
                                             existing_package_dict['name'])
 
+                    for field in p.toolkit.aslist(config.get('ckan.harvest.not_overwrite_fields')):
+                        if field in existing_package_dict:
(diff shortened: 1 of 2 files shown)


ckanext/harvest/harvesters/base.py
```diff
@@ -16,6 +16,7 @@
 from ckan.model import Session, Package, PACKAGE_NAME_MAX_LENGTH
 
 from ckan.logic.schema import default_create_package_schema
+from ckan.lib.base import config
```
amercader, line 19: `config` is already imported in line 12

amercader (approved):
This looks good @frafra , see comments. Also can you document this in the README please? Thanks

frafra:
@amercader Thanks for the review. Fixes and improvements applied.

PR #484, opened 2021-12-21
Add option keep-current to clearsource_history command

seitenbau-govdata:
Based on #397 we added the option to keep the latest related objects to a harvest source when running clearsource_history.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -242,13 +242,15 @@ The following operations can be run from the command line as described underneat
         - clears all datasets, jobs and objects related to a harvest source,
           but keeps the source itself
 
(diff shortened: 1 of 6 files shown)


metaodi:
I think this option should be called "keep-current" instead of "keep-actual", since you want to keep the current active object. WDYT?

seitenbau-govdata:
@metaodi Yes, I think same. Our intention was changing as less as possible from the pull request this one is based on. But you have convinced me to make the changes. If nobody else wants to keep the current naming of the variable we will rename it to `keep-current`.

PR #487, opened 2022-01-18
Futureproof version check to work also on 3.0+

Zharktas:
Coworker of mine noticed that version check would fail when the ckan version 3.0 is released as the minor version is no longer greater than 9, this updated check will also work on 3.0 and any patch releases which can be defined in the check.

### DIFF
--- a/ckanext/harvest/templates/base.html
+++ b/ckanext/harvest/templates/base.html
@@ -3,7 +3,7 @@
 {% block styles %}
     {{ super() }}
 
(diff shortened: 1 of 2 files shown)


ThrawnCA:
Several templates were overlooked:

`ckanext/harvest/templates/snippets/source_item.html`
`ckanext/harvest/templates/source/read_base.html`
`ckanext/harvest/templates/source/source_item.html`
`ckanext/harvest/templates/source/admin_base.html`

PR #490, opened 2022-03-04
Add BS5 support

TomeCirun:
Bootstrap 5 has changed the naming of the data attributes that are generally used by its components that use JavaScript as part of their functionality.

Previously, most components that relied on some JavaScript functionalities would have data attributes starting with data-.
In Bootstrap 5, data attributes for these components now start with data-bs to easily namespace Bootstrap attributes.


ckanext/harvest/templates/source/new_source_form.html
```diff
@@ -37,7 +37,7 @@
         <label class="radio">
           <input type="radio" name="source_type" value="{{ harvester['name'] }}" {{ "checked " if checked }} data-module="harvest-type-change">
           {{ harvester['title'] }}
-          <i class="fa fa-question-circle icon-question-sign muted" title="{{ harvester['description'] }}" data-toggle="tooltip"></i>
+          <i class="fa fa-question-circle icon-question-sign muted" title="{{ harvester['description'] }}" data-bs-toggle="tooltip"></i>
```
amercader, line 40: If the only changes needed to support BS5 are these two small data attributes I think it's overkill to duplicate the whole templates directory. Let's just duplicate the data attributes with the BS3 and BS5 versions and the same template can be used in both versions (I assume each version will ignore the other data attribute):

```suggestion
          <i class="fa fa-question-circle icon-question-sign muted" title="{{ harvester['description'] }}" data-bs-toggle="tooltip" data-toggle="tooltip"></i>
```

ckanext/harvest/plugin/__init__.py
```diff
@@ -244,7 +244,10 @@ def update_config(self, config):
             log.warn('Old genshi templates not supported any more by '
                      'ckanext-harvest so you should set ckan.legacy_templates '
                      'option to True any more.')
-        p.toolkit.add_template_directory(config, '../templates')
+
```
amercader, line 247: These changes are not needed any more

amercader:
Thanks @TomeCirun

PR #492, opened 2022-03-22
Add support for CKAN 2.10

TomeCirun:
I extended the main plugin `Harvest` to support `CKAN 2.10`


ckanext/harvest/plugin/__init__.py
```diff
@@ -29,6 +29,8 @@ class DefaultTranslation():
 
 if p.toolkit.check_ckan_version(min_version='2.9.0'):
     from ckanext.harvest.plugin.flask_plugin import MixinPlugin
+elif p.toolkit.check_ckan_version(min_version='2.10.0'):
```
amercader, line 32: We can't do that @TomeCirun, otherwise we will miss all the methods that `flask_plugin.MixinPlugin` implements. If we wanted to add another mixin we would need to import a separate mixin class on CKAN >= 2.10 and use it on the `Harvest` class:

```python
class Harvest(MixinPlugin, MixinPlugin2, p.SingletonPlugin, DefaultDatasetForm, DefaultTranslation):
```

But let's not do that. Let's just keep all the changes in this file and implement both hooks for CKAN pre and post 2.10. As the function signatures haven't changed we can just reference the new hooks from the old ones. Something like:

```python 
class Harvest(MixinPlugin, p.SingletonPlugin, DefaultDatasetForm, DefaultTranslation):
    # ...
    p.implements(p.IPackageController, inherit=True)

    # IPackageController

    # CKAN < 2.10 hooks
    def after_create(self, context, data_dict):
        self.after_dataset_create(context, data_dict)

    def after_update(self, context, data_dict):
        self.after_update_update(context, data_dict)
   # etc...

   # CKAN >= 2.10 hooks

   def after_dataset_create(self, context, data_dict):
       if 'type' in data_dict and data_dict['type'] == DATASET_TYPE_NAME and not self.startup:
            # Create an actual HarvestSource object
            _create_harvest_source_object(context, data_dict)

   def after_dataset_update(self, context, data_dict):
       if 'type' in data_dict and data_dict['type'] == DATASET_TYPE_NAME:
            # Edit the actual HarvestSource object
            _update_harvest_source_object(context, data_dict)

   # etc...

```
Does this make sense @TomeCirun ?

TomeCirun: Yep, that makes sense, thanks for your help @amercader
At first, I was thinking to make it in the same file and put comments for each, but then I ended up with a separate plugin and forgot about method implementations...

ckanext/harvest/plugin/__init__.py
```diff
@@ -166,6 +169,120 @@ def after_show(self, context, data_dict):
 
         return data_dict
 
+    # CKAN >= 2.10 hooks
+    def after_dataset_create(self, context, data_dict):
```
amercader, line 173: You are still duplicating the same logic in `after_create()` and `after_dataset_create()`. The idea is that the actual logic is located in the hook that will be supported going forward (`after_dataset_create()`) and that `after_create()` is just a call to it. Check my example in the previous comment:

```python
    # CKAN < 2.10 hooks
    def after_create(self, context, data_dict):
        return self.after_dataset_create(context, data_dict)
```

TomeCirun:
@amercader i think this is ready to be merged

amercader:
Thanks @TomeCirun, would you mind enabling the 2.10 tests as described in https://github.com/ckan/ckan/issues/6787 in a separate PR? There's still work needed for full 2.10 support

TomeCirun:
> Thanks @TomeCirun, would you mind enabling the 2.10 tests as described in [ckan/ckan#6787](https://github.com/ckan/ckan/issues/6787) in a separate PR? There's still work needed for full 2.10 support

Yep, i can do that.

PR #500, opened 2022-05-09
Add the harvesters tab to ckan admin dashboard

tino097:
Add the `harvesters` tab for easier access as sysadmin though the `/harvest` endpoint is not behind authorization check.


seitenbau-govdata:
Fixes https://github.com/ckan/ckanext-harvest/issues/547

Issue #293, opened 2017-06-20
Option to clear only old history entries


torfsen:
Currently we have the `clearsource_history` paster command for deleting old jobs but keeping the sources and the datasets. This is nice, but would be even better if we could use it to only delete the older jobs and keep more recent ones, e.g. something like

    paster clearsource_history --older-than 30d

to remove all jobs older than 30 days.

pduchesne:
Is this planned ?

I think this is more than nice-to-have.
I had to clear history for a harvest source (db got bloated because of the many harvest objects). But by doing so, all harvest objects being gone, the harvester (CSW in my case) could no longer find its previous harvested objects, and re-harvested the whole source. And duplicated the data.

So I think clearsource_history should at least preserve the most recent/currently active harvest object.

seitenbau-govdata:
Related PR: https://github.com/ckan/ckanext-harvest/pull/484
With the new option `-k true` (with the CKAN click command) the latest harvest jobs with the current harvest objects will be preserved.

PR #503, opened 2022-05-18
Allow full clear of completed jobs

bonnland:
The current behavior of the "source clear-history" command is to retain all job ids where some harvested dataset came from that job id.    This seems unnecessary, and for our CKAN instance, with many thousands of harvested datasets that get updated in small batches, it means that there is always a very long job history list.   

This PR will require a small 2-line change to the ckan/ckanext-spatial WAF harvesting code, to handle the case where a harvest job ID has been cleared.    It is possible that other harvesters will need small adjusts like this; I've only tested the WAF harvesting code because this is my organization's particular use case.

I have also made a suggested change to the options for this command, so that "keep currently running jobs" is always done.   It seems potentially confusing, possibly unhelpful, and unnecessary to clear the history of jobs that are still running.

I know that this initial PR version has failing tests.    I do not have testing set up on my current vagrant VM instance, and will try to fix the tests by looking at the Github tests output.  

Please feel free to suggest possible alternatives or edits.     Thank you!

### DIFF
--- a/ckanext/harvest/model/__init__.py
+++ b/ckanext/harvest/model/__init__.py
@@ -28,6 +28,7 @@ __all__ = [
     'HarvestObject', 'harvest_object_table',
     'HarvestGatherError', 'harvest_gather_error_table',
     'HarvestObjectError', 'harvest_object_error_table',
+    'HarvestObjectExtra', 'harvest_object_extra_table',
     'HarvestLog', 'harvest_log_table'
 ]
 
@@ -356,7 +357,7 @@ def define_harvester_tables():
         Column('state', types.UnicodeText, default=u'WAITING'),
         Column('metadata_modified_date', types.DateTime),
         Column('retry_times', types.Integer, default=0),
-        Column('harvest_job_id', types.UnicodeText, ForeignKey('harvest_job.id')),
+        Column('harvest_job_id', types.UnicodeText, ForeignKey('harvest_job.id', ondelete='SET NULL'), nullable=True),
         Column('harvest_source_id', types.UnicodeText, ForeignKey('harvest_source.id')),
         Column('package_id', types.UnicodeText, ForeignKey('package.id', deferrable=True),
                nullable=True),
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ node_modules
 *.project
 .eggs
 .vscode/
+.idea/
--- a/ckanext/harvest/logic/action/update.py
+++ b/ckanext/harvest/logic/action/update.py
@@ -313,15 +313,13 @@ def harvest_abort_failed_jobs(context, data_dict):
 
 def harvest_sources_job_history_clear(context, data_dict):
     '''
-    Clears the history for all active harvest sources. All jobs and objects related to a harvest source will
-    be cleared, but keeps the source itself.
+    Clears the history for all active harvest sources. All jobs and objects related to completed harvests will
+    be cleared, but keeps the source itself and the most recently harvested objects.
     This is useful to clean history of long running harvest sources to start again fresh.
     The datasets imported from the harvest source will NOT be deleted!!!
 
     '''
     check_access('harvest_sources_clear', context, data_dict)
-
-    keep_current = data_dict.get('keep_current', False)
 
     job_history_clear_results = []
     # We assume that the maximum of 1000 (hard limit) rows should be enough
@@ -331,7 +329,7 @@ def harvest_sources_job_history_clear(context, data_dict):
         for data_dict in harvest_packages:
             try:
                 clear_result = get_action('harvest_source_job_history_clear')(
-                    context, {'id': data_dict['id'], 'keep_current': keep_current})
+                    context, {'id': data_dict['id']})
                 job_history_clear_results.append(clear_result)
             except NotFound:
                 # Ignoring not existent harvest sources because of a possibly corrupt search index
@@ -343,7 +341,7 @@ def harvest_sources_job_history_clear(context, data_dict):
 
 def harvest_source_job_history_clear(context, data_dict):
     '''
-    Clears all jobs and objects related to a harvest source, but keeps the source itself.
+    Clears all jobs and objects related to a harvest source's completed jobs.
     This is useful to clean history of long running harvest sources to start again fresh.
     The datasets imported from the harvest source will NOT be deleted!!!
 
@@ -354,7 +352,6 @@ def harvest_source_job_history_clear(context, data_dict):
     check_access('harvest_source_clear', context, data_dict)
 
     harvest_source_id = data_dict.get('id', None)
-    keep_current = data_dict.get('keep_current', False)
 
     source = HarvestSource.get(harvest_source_id)
     if not source:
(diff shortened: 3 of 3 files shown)


bonnland:
Related PR: https://github.com/ckan/ckanext-spatial/pull/284

bonnland:
Note:   I've verified on my vagrant VM that currently running harvest jobs are not cleared when the "job history clear" command is given.   It is a source-based CKAN 2.9.5 install running uwsgi+nginx, with WAF harvesting enabled.

bonnland:
Related to #293

ckanext/harvest/model/__init__.py
```diff
@@ -356,7 +357,7 @@ def define_harvester_tables():
         Column('state', types.UnicodeText, default=u'WAITING'),
         Column('metadata_modified_date', types.DateTime),
         Column('retry_times', types.Integer, default=0),
-        Column('harvest_job_id', types.UnicodeText, ForeignKey('harvest_job.id')),
+        Column('harvest_job_id', types.UnicodeText, ForeignKey('harvest_job.id', ondelete='SET NULL'), nullable=True),
```
Zharktas, line 360: Changing table structure would require migration script, otherwise it would not be applied to any instance.

seitenbau-govdata:
Hi @bonnland. Thanks for the interesting pull request. Could you explain the advantage of deleting the harvest jobs and keeping the harvest objects?

The current implementation with the option `-k true` preserves the harvest jobs with current harvest objects as long as a harvest job has at least one current harvest object. And these harvest jobs with their reports are still available in the UI.

bonnland:
I just realized that the user option `--keep_current` for keeping current harvest objects would no longer be needed with this pull request.   

If there are organizations who want duplicate packages in the package table, this pull request should not be accepted.   It's hard to think of a reason why duplicate records would be good, though, from my perspective.

bonnland:
> Hi @bonnland. Thanks for the interesting pull request. Could you explain the advantage of deleting the harvest jobs and keeping the harvest objects?
> 

Hi @seitenbau-govdata, thanks for the attention and interest.   Keeping the harvest objects allows tracking of what has been harvested already.   If all harvest objects are cleared as part of the history clear, then the harvester will re-harvest datasets that already appear in the package table, and duplicate rows in the package table will be created.    For our organization, this is costly because our harvest sources have thousands of datasets.   The package table fills up quickly with duplicate rows if harvest objects are not kept, and it can take hours to re-harvest everything.

ckanext/harvest/model/__init__.py
bonnland: I agree that a schema change creates some challenges for existing databases.   However, the `ckanapi` package gives very nice support for exporting and re-importing users, packages, and organizations, so that a fresh database instance with imported users, etc is not very time consuming.     I could write up a set of instructions for the Wiki if this is useful.

bonnland:
> And these harvest jobs with their reports are still available in the UI.

For our purposes, the harvest job reports help in the short term, to fix harvesting errors.   Their usefulness rarely goes past a few days. 

Is it useful to have job reports for cases where the datasets were harvested successfully?   Because it seems that harvest job ids are kept when datasets are harvested successfully.    If you can explain how this is helpful, I would appreciate knowing the advantages.

ckanext/harvest/utils.py
```diff
@@ -216,18 +216,12 @@ def clear_harvest_source_history(source_id, keep_current):
     if source_id is not None:
         tk.get_action("harvest_source_job_history_clear")(context, {
             "id": source_id,
-            "keep_current": keep_current
-        })
+            })
         return "Cleared job history of harvest source: {0}".format(source_id)
```
bonnland, line 220: Note the language used here.   This command is most useful if it clears the "job history", not the entire source history.    Perhaps a change in the command name to "clear-job-history" would be clearer.

bonnland:
I should probably not forget to add that our organization adds new records individually, at an average rate of 3-4 per week. That means I have a "cleared" job history list with over 100 entries for some of our WAFs.   This might not be the usual case for others, so the urgency for this PR is probably not as relevant for others as it might be for us.

Zharktas:
> > And these harvest jobs with their reports are still available in the UI.
> 
> For our purposes, the harvest job reports help in the short term, to fix harvesting errors. Their usefulness rarely goes past a few days.
> 
> Is it useful to have job reports for cases where the datasets were harvested successfully? Because it seems that harvest job ids are kept when datasets are harvested successfully. If you can explain how this is helpful, I would appreciate knowing the advantages.

Not all instances harvest daily. For you the history might be relevant only for few days, but other might have a need for what happened three months ago.

What I would do is have separate command for what you are trying to achieve or at least an option in the current one. It would be a nasty surprise for someone running this command and noticing the functionality has changed.

bonnland:
> What I would do is have separate command for what you are trying to achieve or at least an option in the current one. It would be a nasty surprise for someone running this command and noticing the functionality has changed.

I am not sure I understand.   The command is run when someone wants to clear the job history.    If they don't want to clear the job history, then they do not run the command.

What this pull request does is make this command as it did before 2016 or 2017.   I am not sure why it was changed; it worked very well before.

Zharktas:
There's discussion on https://github.com/ckan/ckanext-harvest/pull/484 and https://github.com/ckan/ckanext-harvest/pull/397 why the change was made originally.

bonnland:
> There's discussion on #484 and #397 why the change was made originally.

OK, perhaps it would be better if a new command option is made available.   How does "harvester source clear-job-history" sound?    It would still require a change to the foreign key constraint on the harvest_job table.

seitenbau-govdata:
> There used to be a command that would clear all completed jobs. It was very helpful. I suppose it disappeared when the `clearsource_history` command was added.

No, there wasn't removed any command when adding the command `clearsource_history`. The only command was and still is the `clearsource` (as click command `clear source`) which deletes the source with all datasets. That was the reason why we introduced the `clearsurce_history` command, because there was no command like deleting only the harvest job history. If there had exists such a command we didn't introduces the `clearsurce_history` command.

bonnland:
> No, there wasn't removed any command when adding the command `clearsource_history`. The only command was and still is the `clearsource` (as click command `clear source`) which deletes the source with all datasets. That was the reason why we introduced the `clearsurce_history` command, because there was no command like deleting only the harvest job history. If there had exists such a command we didn't introduces the `clearsurce_history` command.

I remember a time when I could clear all completed jobs, and it would not cause duplicate rows in the package table after it was used.    This issue of the package table filling up is potentially serious, and I am surprised it has not come up somewhere before.

seitenbau-govdata:
> I remember a time when I could clear all completed jobs, and it would not cause duplicate rows in the package table after it was used. This issue of the package table filling up is potentially serious, and I am surprised it has not come up somewhere before.

Maybe somewhere else in a fork? But unfortunately not in `ckanext-dcat`. Yes, I agree. With many thousands of datasets it is really serious and a huge problem. After going into production in early 2016 with more than 20.000 datasets and a harvesting interval of 2 days we pointed out after a few month that the size of our database was increasing and the harvest UI was getting slower. Therefore, we started to implement the new command.

bonnland:
> With many thousands of datasets it is really serious and a huge problem.

It is helpful to know that our organization is not the only one who has had this problem.   The current interface does not provide any way of clearing past jobs without creating an entire new set of duplicate rows in the package table.

seitenbau-govdata:
> It is helpful to know that our organization is not the only one who has had this problem. The current interface does not provide any way of clearing past jobs without creating an entire new set of duplicate rows in the package table.

Actually this should be possible with the new option `--keep-current` respectively `-k`. We use `harvester source clear-history -k true` to delete the old harvest jobs and keep the latest harvest jobs with the current harvest objects. Does this not work for you?

bonnland:
> We use `harvester source clear-history -k true` to delete the old harvest jobs and keep the latest harvest jobs with the current harvest objects.

Do you find the remaining job reports useful in any way?

bonnland:
If the `keep_current` behavior is useful to some organizations, there are a few ways we could go that would also work for our organization:  Here are a couple of possible changes to the current behavior that would work for us:

- When `keep_current=false`, also delete and purge harvested datasets in the package table.  This would also prevent duplicate entries from being created in the package table over the long term.   But then it would be very similar to the behavior of the "clear source" command, and it would make datasets disappear from CKAN.
- When `keep_current=false`, do not delete all information about what has been harvested.    Retain all harvest objects that are still current, but clear the job_id field. This would prevent duplicate entries in the package table by preserving the prior state of harvesting for the source.  
- Add a new command "clear-job-history" that removes past jobs without losing information about what has been harvested already.   This would also prevent duplicate entries in the package table. 

I could be wrong, but it seems that the second and third choices would require that the "job_id" field in the harvest_object table can be set to NULL.

bonnland:
> Maybe somewhere else in a fork?

You may be correct that `clearsource_history` without `keep_current=true` has always led to the package table growing over time.  Perhaps it was harder to notice before because the harvested dataset URLs did not change on every re-harvest after the command was used.

Issue #502, opened 2022-05-10
Broken unicode support for dataset search 


jan-imrich:
We're using CKAN 2.8.10 (legacy python) in Docker. Latest release of ckanext-harvest.
Searching for datasets by tag - for example /dataset?tags=abcčd yields server error:

```
Error - <type 'exceptions.UnicodeEncodeError'>: 'ascii' codec can't encode character u'\u010d' in position 9: ordinal not in range(128)
URL: http://.../dataset?tags=abc%C4%8Dd
File '/usr/lib/ckan/venv/local/lib/python2.7/site-packages/weberror/errormiddleware.py', line 171 in __call__
  app_iter = self.application(environ, sr_checker)
File '/usr/lib/ckan/venv/local/lib/python2.7/site-packages/webob/dec.py', line 147 in __call__
  resp = self.call_func(req, *args, **self.kwargs)
File '/usr/lib/ckan/venv/local/lib/python2.7/site-packages/webob/dec.py', line 208 in call_func
  return self.func(req, *args, **kwargs)
File '/usr/lib/ckan/venv/local/lib/python2.7/site-packages/fanstatic/publisher.py', line 234 in __call__
  return request.get_response(self.app)
File '/usr/lib/ckan/venv/local/lib/python2.7/site-packages/webob/request.py', line 1053 in get_response
  application, catch_exc_info=False)
File '/usr/lib/ckan/venv/local/lib/python2.7/site-packages/webob/request.py', line 1022 in call_application
  app_iter = application(self.environ, start_response)
File '/usr/lib/ckan/venv/local/lib/python2.7/site-packages/webob/dec.py', line 147 in __call__
  resp = self.call_func(req, *args, **self.kwargs)
File '/usr/lib/ckan/venv/local/lib/python2.7/site-packages/webob/dec.py', line 208 in call_func
  return self.func(req, *args, **kwargs)
File '/usr/lib/ckan/venv/local/lib/python2.7/site-packages/fanstatic/injector.py', line 54 in __call__
  response = request.get_response(self.app)
File '/usr/lib/ckan/venv/local/lib/python2.7/site-packages/webob/request.py', line 1053 in get_response
  application, catch_exc_info=False)
File '/usr/lib/ckan/venv/local/lib/python2.7/site-packages/webob/request.py', line 1022 in call_application
  app_iter = application(self.environ, start_response)
[...]

PR #508, opened 2022-07-14
Fixes unicode error in Python 2

seitenbau-govdata:
Fixes https://github.com/ckan/ckanext-harvest/issues/502

### DIFF
--- a/ckanext/harvest/plugin/__init__.py
+++ b/ckanext/harvest/plugin/__init__.py
@@ -2,6 +2,7 @@
 
 import os
 import json
+import six
 from logging import getLogger
 
 from six import string_types
@@ -110,7 +111,7 @@ class Harvest(MixinPlugin, p.SingletonPlugin, DefaultDatasetForm, DefaultTransla
 
         fq = search_params.get("fq", "")
         if "dataset_type:harvest" not in fq:
-            fq = "{0} -dataset_type:harvest".format(search_params.get("fq", ""))
+            fq = "{0} -dataset_type:harvest".format(fq.encode('utf8') if six.PY2 else fq)
             search_params.update({"fq": fq})
 
         return search_params


amercader:
@seitenbau-govdata this should do the trick:

```diff
diff --git a/ckanext/harvest/tests/test_blueprint.py b/ckanext/harvest/tests/test_blueprint.py
index dbbca7d..a6bef99 100644
--- a/ckanext/harvest/tests/test_blueprint.py
+++ b/ckanext/harvest/tests/test_blueprint.py
@@ -1,3 +1,4 @@
+# coding: utf-8
 import six
 import pytest
 
@@ -126,3 +127,15 @@ class TestBlueprint():
         response = app.get(url, extra_environ=env)
 
         _assert_in_body(job['id'], response)
+
+    def test_search_dataset_special_characters(self, app):
+
+        value = u"abcčd"
+
+        dataset = factories.Dataset(tags=[{"name": value}])
+
+        url = u"/dataset?tags=" + value
+
+        response = app.get(url)
+
+        _assert_in_body(dataset["title"], response)

```

Note I didn't use `url_for` because there was another bug in CKAN core with unicode handling but went too down the rabbit hole.

The 2.10 failures should be fixed now

seitenbau-govdata:
@amercader I have tried the quite similar test.
```
def test_package_search_interface_before_search_page_is_rendered(self, app):

    dataset1 = factories.Dataset(title=u'Fïrst dataset')

    response = app.get(u'/dataset?title=Fïrst')

    _assert_in_body(dataset1['title'], response)
```

But it hasn't worked because of a unicode error in `app.get(u'/dataset?title=Fïrst')`.

And I was getting into the rabbit whole with using `url_for`. 😃 I get it to work with a custom `url_for` implementation (e.g. as in ckanext-dcat) that switches to the old style if it is CKAN < 2.9. But then the test didn't fail without the fix as I would expect. Probably because there is some other unicode encoding in the `url_for` core code.

seitenbau-govdata:
@amercader I have removed the test now.

PR #512, opened 2022-09-24
Allow purge the harvest source

LittleRed945:
Make a new list item that can purge the harvest source so that one can directly delete the data in the harvest source and package table when a harvest source is no longer needed.


LittleRed945:
Does anyone have any suggestions to solve the ckan2.8 failure problem?

PR #516, opened 2022-12-04
Add CSRF Protection Support

### LINKED ISSUE #511: CSRF protection breaks harvest source creation form
Using harvest with the latest CKAN master returns 400 when attempting to create a new harvest source via the web UI form.

image

pdekraker-epa:
Fixes #511

### DIFF
--- a/ckanext/harvest/templates/source/new_source_form.html
+++ b/ckanext/harvest/templates/source/new_source_form.html
@@ -14,7 +14,7 @@
       </span>
   {% endcall %}
 
-  {{ h.csrf_input() }}
+{{ h.csrf_input() if 'csrf_input' in h }}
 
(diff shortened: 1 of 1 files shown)


ckanext/harvest/templates/source/new_source_form.html
```diff
@@ -14,6 +14,8 @@
       </span>
   {% endcall %}
 
+  {{ h.csrf_input() }}
```
amercader, line 17: Thansk @pdekraker-epa! 

This will fail in older CKAN versions that don't have this helper available, so we need:

```suggestion
{{ h.csrf_input() if 'csrf_input' in h }}
```

Also can you merge the latest master branch? That should fix the tests

PR #517, opened 2022-12-21
added unescape for email text body to avoid encoded characters

Jin-Sun-tts:
The harvest report emails sent to users contains encoded characters, like `&#39;` and `&#34;`

This makes it hard for users to understand the error message. For example:

before 
`ERROR #1: &#39;description&#39;:&#39;&#39; is too short.`
 
after
`ERROR #1: 'description':'' is too short.`

### DIFF
--- a/ckanext/harvest/logic/action/update.py
+++ b/ckanext/harvest/logic/action/update.py
@@ -804,7 +804,7 @@ def prepare_error_mail(context, source_id, status):
     elif six.PY2:
         import HTMLParser
         body = HTMLParser.HTMLParser().unescape(body)
-    
+
     subject = '{} - Harvesting Job - Error Notification'\
         .format(config.get('ckan.site_title'))
 
--- a/ckanext/harvest/logic/action/update.py
+++ b/ckanext/harvest/logic/action/update.py
@@ -3,7 +3,6 @@
 import hashlib
 import json
 import six
-import html
 
 import logging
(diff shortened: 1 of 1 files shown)


jbrown-xentity:
cc @amercader @avdata99 , this is due to data.gov email formatting problems.

avdata99:
@jbrown-xentity , I think it's good for Python 3 but we need to cover Python 2 users (CKAN <= 2.9.py2) users
`html.unescape` is only available from Python 3.4
Tests for this extension still check from CKAN 2.7

jbrown-xentity:
Thanks @avdata99 , we'll push something with `six` handling the Py2 use case...
Any reason why the checks don't show up on the PR? When we push more commits, is there a way for us to know the results of the tests  without asking you?

avdata99:
> Any reason why the checks don't show up on the PR? When we push more commits, is there a way for us to know the results of the tests without asking you?

@jbrown-xentity, this repository takes care of the minutes to execute the GitHub actions. This is why it requires the authorization of a repository administrator user to run.
Forks of this repository do execute these actions (including tests) and each external user can check their own tests.

avdata99 (approved):
LGTM

jbrown-xentity:
@amercader  can you take a look?

jbrown-xentity:
@amercader ping?

PR #520, opened 2023-02-28
Drop support old versions

pdelboca:
This PR deprecates support for 2.7 and 2.8 to match our policy: https://docs.ckan.org/en/latest/maintaining/releases.html#supported-versions

## Breaking changes
 - `h.bootstrap_version()` no longer exist since it is no longer needed to inject CSS classes
 - Support for old Pylons route names has been removed. Example: calling `url_for("harvest_read")` will no longer work. URLs for `ckanext-harvest` needs to respect Flask's syntax (`url_for("harvest.read")`, etc)


amercader:
Can you resolve the conflicts @pdelboca ?

PR #521, opened 2023-03-09
Do not duplicate harvest_extras if exist in root schema

pdelboca:
This PR refactors and document the logic of our `dataset_before_index` method.

With the new changes, harvest metadata will not be added to extras if it exist in the root schema of the dataset. This will allow other extensions or implementations to add harvest metadata to the main package schema without getting duplicated errors when trying to update or patch the dataset.

### DIFF
--- a/ckanext/harvest/plugin/__init__.py
+++ b/ckanext/harvest/plugin/__init__.py
@@ -116,7 +116,37 @@ class Harvest(MixinPlugin, p.SingletonPlugin, DefaultDatasetForm, DefaultTransla
 
         return search_params
 
+    def _add_or_update_harvest_metadata(self, key, value, data_dict):
+        """Adds extras fields or updates them if already exist."""
+        if not data_dict.get("extras"):
+            data_dict["extras"] = []
+
+        for e in data_dict.get("extras"):
+            if e.get("key") == key:
+                e.update({"value": value})
+                break
+        else:
+            data_dict["extras"].append({"key": key, "value": value})
+
     def before_dataset_index(self, pkg_dict):
+        """Adds harvest metadata to the extra field of the dataset.
+
+        This method will add or update harvest related metadata in `pkg_dict`,
+        `data_dict` and `validated_data_dict` so it can be obtained when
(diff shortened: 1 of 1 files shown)


ckanext/harvest/plugin/__init__.py
```diff
@@ -125,59 +155,37 @@ def before_dataset_index(self, pkg_dict):
...
+        validated_data_dict = json.loads(pkg_dict["validated_data_dict"])
+        for key, value in harvest_extras:
+            if key in validated_data_dict.keys():
+                validated_data_dict[key] = value
+                continue
+            self._add_or_update_harvest_metadata(key, value, validated_data_dict)
+
+        # Add harvest extras to main indexed pkg_dict
+        for key, value in harvest_extras:
+            pkg_dict[key] = value
```
pdelboca, line 185: This code has been here but I don't see it working. 

Indexed packages do not have this attributes when calling Solr directly.

pdelboca:
@amercader As a continuation of this PR I want to add some documentation about `pkg_dict`, `data_dict` and `validated_data_dict` on CKAN Docs.

I still don't get quite well the uses and difference betwen them.

ckanext/harvest/plugin/__init__.py
amercader: This was done to add the harvest field as the ["catch-all" string field](https://github.com/ckan/ckan/blob/98d2164f48efff6d2448acf3ba5dc0c8c0796fdc/ckan/config/solr/schema.xml#L175) in Solr. These values are indexed but not stored, so that's why you don't see them when calling Solr directly.
With your changes, we should only add them if not already there in `pkg_dict`

amercader:
@pdelboca This looks good!

re the different dicts:
* `pkg_dict` means "the dataset dict that will be sent to Solr for indexing"
* `data_dict` means the default dataset metadata dict without any validation applied (eg all custom fields are extras)
* `validated_data_dict` means the dataset metadata dict after applying validation (eg what you get on `package_show` with the schema customizations applied)

PR #533, opened 2023-05-26
Use tuples for dictization of harvest job errors

seitenbau-govdata:
In #529 we changed the dictization of harvest errors. While this solves the problem #528 it causes a problem with displaying the error messages in the job overview.
Therefore we would fix this by using a tuple to restore the old behavior. 

**Details**
Since the update to `sqlalchemy 1.4.x` the harvest error objects returned by the database in https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/logic/dictization.py#L79 are no longer serializable which causes commands like `harvester reindex` or `harvester clearsource_history` to fail if there are any harvest jobs with errors. In #529 these harvest error objects are transformed into a dict to fix this problem. 
By doing so this data can no longer be accessed by index but this is expected by the HTML template (https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/templates/snippets/job_error_summary.html) and potentially by further CKAN extensions.
So our proposed solution would be to use a tuple for the data to regain the old behavior and still fix the serialization problems.

### DIFF
--- a/ckanext/harvest/templates/snippets/job_error_summary.html
+++ b/ckanext/harvest/templates/snippets/job_error_summary.html
@@ -1,7 +1,7 @@
 {#
 Displays a table with a summary of the most common errors for a job
 
-error_summary        - List of tuples with (message, count)
+error_summary        - List of dicts with message and error_count
 
 Example:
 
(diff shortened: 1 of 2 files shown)


seitenbau-govdata:
Could somebody take a quick look if this is a viable solution? Maybe @amercader or @pdelboca

amercader:
@seitenbau-govdata I think it's very unlikely that other extensions are using this object so I'd rather update the templates to use a dict than return a tuple. Dicts are the preferred output format for the API anyway

seitenbau-govdata:
@amercader Thanks for the feedback.
We've made a new commit with changes to the template.

PR #534, opened 2023-06-23
fix url endpoint for job_show

Jin-Sun-tts:
Fix URL endpoints for job_show: from `harvest.job_show` to `harvester.job_show`.


seitenbau-govdata:
Thanks for the contribution @Jin-Sun-tts . LGTM, but could you remove the commit bumping the version in `setup.py`? The version should not be increased within in a pull request changing functionality. I will release a new version once after the the pull request is merged.

seitenbau-govdata:
Thanks @Jin-Sun-tts

PR #540, opened 2023-11-02
feat: switch to alembic migrations

smotornyuk:
@amercader , I mentioned this change at the dev meeting

To make this extension compatible with CKAN after merging sqlalchemy-2.0 PR, at least we have to pass [`engine` to all `create` and `exists` calls](https://github.com/ckan/ckanext-harvest/blob/master/ckanext/harvest/model/__init__.py#L50-L64)

But, if ckanext-harvest targets only CKAN >= v2.9, we can switch to alembic migrations. When I had done it, I decided to move all table definitions to the top level of the module.

Finally, I created a second migration. It adds PostgreSQL level `ON DELETE CASCADE` to all the foreign keys. It's not required for sqlalchemy upgrade, but it really simplifies cleaning the tables manually


pdelboca:
@smotornyuk I like this one. 

I have tested it locally agains the `SQLAlchemy 2.0` branch and it is working. I executed a `ckan_harvester` against a public instance and I found no issues.

PR #557, opened 2024-11-15
Make user agent string customisable

ThrawnCA:
Some servers whitelist or blacklist specific user agents, so allow the header value to be configured.

Also fix the build on latest `ckan-dev` container.


seitenbau-govdata:
@ThrawnCA Thanks for the contribution. LGTM. Maybe it would be helpful if the configuration value is validated that it is a simple string to avoid problems sending the request.

PR #558, opened 2024-12-20
Update manifest to include alembic configuration

tino097:
Update manifest to include alembic configuration

similar issue as https://github.com/ckan/ckan/pull/8510


avdata99 (approved):
By mistake I created #560 with the same change
This is required

PR #560, opened 2025-01-09
Ensure we install migrations

avdata99:
We started migrations for harvest 1.6.0 (CKAN 2.11) but e forgot to add migrations


avdata99:
Duplicated from #558

Issue #565, opened 2025-03-08
Error when clearing harvest source: relation "package_ids_to_delete" already exists


FuhuXia:
CKAN core 2.11, ckanext-harvest v1.6.1, postgresql 15.10

When you clear a harvest source, first time it is fine. But when you try to clear source again, either on the same source or another source, you see error:
```
(psycopg2.errors.DuplicateTable) relation "package_ids_to_delete" already exists ...
```

PR #566, opened 2025-03-08
drop temp table package_ids_to_delete

FuhuXia:
For https://github.com/ckan/ckanext-harvest/issues/565.
This change is to avoid error `relation "package_ids_to_delete" already exists`

This error is happening because you're trying to create a temporary table that already exists in your current session.
We have three ways to fix it:
- `DROP TABLE IF EXISTS package_ids_to_delete;`
- `CREATE TEMP TABLE IF NOT EXISTS package_ids_to_delete AS (...)`
- `CREATE TEMP TABLE package_ids_to_delete ON COMMIT DROP AS (...)`

The first one is chosen since it is more clean and straightforward.

### DIFF
--- a/ckanext/harvest/logic/action/update.py
+++ b/ckanext/harvest/logic/action/update.py
@@ -125,6 +125,7 @@ def harvest_source_clear(context, data_dict):
             WHERE harvest_source_id = '{harvest_source_id}');"""
 
     sql += """
+        DROP TABLE IF EXISTS package_ids_to_delete;
         CREATE TEMP TABLE package_ids_to_delete AS (
             SELECT id FROM package WHERE state = 'to_delete');
         """
--- a/ckanext/harvest/logic/action/update.py
+++ b/ckanext/harvest/logic/action/update.py
@@ -126,7 +126,7 @@ def harvest_source_clear(context, data_dict):
 
     sql += """
         DROP TABLE IF EXISTS package_ids_to_delete;
-        CREATE TEMP TABLE package_ids_to_delete AS (
+        CREATE TEMP TABLE package_ids_to_delete ON COMMIT DROP AS (
             SELECT id FROM package WHERE state = 'to_delete');
         """
 


seitenbau-govdata:
Hi @FuhuXia Thanks for the input. I haven't knew about the temporary tables in the source delete process. 😄 But I imagine that the drop statement could cause problems if there are parallel executions,. So I would prefer the `CREATE TEMP TABLE IF NOT EXISTS` approach.
What Do you think, @amercader ?

amercader:
Yes, `CREATE TEMP TABLE IF NOT EXISTS` seems safer, and in fact we should probably also use `ON COMMIT DROP` to avoid having package ids from different runs

seitenbau-govdata:
I shouldn't have looked at the code so closely. 🙈 
It is needed that the state for the package have to be set to `to_delete`before deleting it from the database?

I think it would be an improvement if  a column for the `harvest_source_id` was added to the temporary table to filter the select statement to it, so several possible parallel executions can be separated and avoids conflicts.

Maybe something similar to that. This is NOT tested!

`CREATE TEMP TABLE package_ids_to_delete AS (SELECT id FROM package WHERE state = 'to_delete', SELECT id AS harvest_source_id FROM harvest_object WHERE harvest_source_id = '{harvest_source_id}'));`

And then use it in the sub-select as follows
`SELECT id FROM package_ids_to_delete WHERE harvest_source_id = '{harvest_source_id}'`

Then it should not be required to drop the temp table on commit. But I have no idea how PostgreSQL handles it when there are parallel executions, but it should certainly safer in any case.

But for the issue this PR addresses, `CREATE TEMP TABLE IF NOT EXISTS` and `ON COMMIT DROP` should work. 😃

FuhuXia:
When you use `CREATE TEMP TABLE package_ids_to_delete AS (...)` in a transaction, here is what will happen:
- The table is invisible to other process until the transaction is finished (either committed or rolled back)
- **After** the transaction is finished, the table stays within the same, and only the same, connection session.

After more thoughts on the choices of 
1. DROP TABLE IF EXISTS package_ids_to_delete;
2. CREATE TEMP TABLE IF NOT EXISTS package_ids_to_delete AS (...)
3. CREATE TEMP TABLE package_ids_to_delete ON COMMIT DROP AS (...)

Choice 1 works. No worries on parallel execution.
Choice 2 has the risk of a following transaction uses an existing table from a previous transaction.
Choice 3 has the risk of a following transaction uses an existing table from a previously rolled-back transaction.
There are might be other choices by mixing these 1/2/3, but we know this table is only used once and only in this code block, Choice 1 is sufficient.

seitenbau-govdata:
You're right. If read the PostgreSQL documentation it runs isolated within the transaction.

```
ON COMMIT

    The behavior of temporary tables at the end of a transaction block can be controlled using ON COMMIT. The three options are:

    PRESERVE ROWS

        No special action is taken at the ends of transactions. This is the default behavior.
    DELETE ROWS

        All rows in the temporary table will be deleted at the end of each transaction block. Essentially, an automatic [TRUNCATE](https://www.postgresql.org/docs/current/sql-truncate.html) is done at each commit. When used on a partitioned table, this is not cascaded to its partitions.
    DROP

        The temporary table will be dropped at the end of the current transaction block. When used on a partitioned table, this action drops its partitions and when used on tables with inheritance children, it drops the dependent children.
```
https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-PARMS-ON-COMMIT

Since`end of a transaction block` can be a `COMMIT` and `ROLLBACK`, I would prefer the `ON COMMIT DROP` because it is a built in feature.

FuhuXia:
Sound good. I added "ON COMMIT DROP", so that the table `package_ids_to_delete` will be gone after transaction is committed.

seitenbau-govdata:
Now it looks like that the `DROP TABLE` isn't necessary anymore. What do you think, @FuhuXia ?

FuhuXia:
@seitenbau-govdata

I think `DROP TABLE` is still needed to deal with edge cases, for example:
- Transaction 1 was executed, but got into trouble half way and rolled back. The temp table stays in the session. `ON COMMIT DROP` could not help in this rollback scenario. 
- Then, transaction 2 in the same session was executed, it saw the temp table. it would raise the same error without the `DROP` phrase.

amercader:
This looks good @seitenbau-govdata feel free to merge

seitenbau-govdata:
@FuhuXia I would expect that the temporary tables in PostgreSQL with the option `ON COMMIT DROP` within a single transaction as used in that case (starts with `BEGIN` and ends with `COMMIT`) is only visible for that current transaction. So different transactions at the same time should not see the other temporary tables and they should have their own copy of the table. But therefore I`m not a database expert, I don`t know exactly how PostgreSQL handles temporary tables in detail. So we can leave the explicitly drop statement.

FuhuXia:
@seitenbau-govdata
What you said is right, different transactions **at the same time** do not see temp tables created in each other. But, a following transaction sees temp tables created by a previously finished transaction of the same session. By finish I mean commit and rollback. The error can be easily replicated by clearing a harvest source twice.
`ON COMMIT DROP` helps in **committed** transaction but does not help in **rolledback** transaction.

